PR #19, opened 2016-12-17
STREAMLINE-600: Handle Byte java type to get avro schema

priyank5485:


schema-registry/serdes/src/main/java/com/hortonworks/registries/schemaregistry/serdes/avro/AvroUtils.java
```diff
@@ -52,7 +52,7 @@ public static Schema getSchemaForPrimitives(Object input) {
             type = Schema.Type.NULL;
         } else if (input instanceof byte[]) {
             type = Schema.Type.BYTES;
-        } else if (input instanceof Integer || input instanceof Short) {
+        } else if (input instanceof Integer || input instanceof Short || input instanceof Byte) {
```
harshach: shouldn't Byte be declared under the else if above?

priyank5485: I dont think so. From what i understand we are trying to assign an avro type for a java type. And Byte should map to int in avro since avro does not have byte. Above if else is for a byte array. Whereas this covers a byte myByte = (byte) 5; statement in java. Correct me if i am wrong. I will change it

PR #21, opened 2016-12-27
Registry 9 Schema creation is allowed with schema name as null

sureshms:


schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
```diff
@@ -327,7 +327,7 @@ private SchemaIdVersion doAddSchemaVersion(String schemaName, SchemaVersion sche
         return new SchemaIdVersion(schemaMetadataInfo.getId(), readEntity(msg, Integer.class));
     }
 
-    private CatalogResponse readCatalogResponse(String msg) {
+    public static CatalogResponse readCatalogResponse(String msg) {
```
harshach: readCatalogResponse is for parsing rest api calls. Users of schemaRegistryClient doesn't really need to understand the api calls it makes do we need to make this public.

sureshms:
I am using the same call here in AvroSchemaRegistryClientTest.java


/** Return true if the schema creation failed as expected */
private void testCreate(SchemaRegistryClient client) {
    boolean failedAsExpected = false;
    SchemaMetadata schemaMetadata = new SchemaMetadata.Builder(name).type(type).schemaGroup(group).
            description("description").build();
    try {
        client.registerSchemaMetadata(schemaMetadata);
    } catch (BadRequestException ex) {
        Response resp = ex.getResponse();
        Assert.assertEquals(test + " - http response unexpected", expectedHttpResponse.getStatusCode(), resp.getStatus());
        CatalogResponse catalogResponse = SchemaRegistryClient.readCatalogResponse(resp.readEntity(String.class));
        Assert.assertEquals(test + " - catalog response unexpected",
                expectedCatalogResponse.getCode(), catalogResponse.getResponseCode());
        failedAsExpected = true;
    }
    Assert.assertTrue(test + " - did not fail as expected", failedAsExpected);
}

PR #26, opened 2017-01-23
Removed declared dependency on 2.10.x Scala so it's possible to crossbuild 2.11 Scala apps via SBT.

orendain:
This version (0.1.0-SNAPSHOT) builds fine.  I've been using this modification for crossbuilding our 2.11 Scala applications.


satishd:
+1

PR #29, opened 2017-01-25
registry-docker

jdye64:
Initial support for building Docker images via Maven. The image can be
built as part of the build process by using the Maven Profile
“-Pdocker” and does expect that “registry-dist” has been previously
built. The easiest way to ensure this is always including “-Pdist” with
the Maven command as well.


jdye64:
I'm not sure of this teams contribution guidelines so I apologize if you want a Github issue or JIRA attached to this let me know and I'll be glad to add one.

harshach:
@jdye64 for the registry we are going with github issues.

PR #33, opened 2017-02-02
Issue#32 Initial version of SchemaRegistryClient with url selectors.

satishd:


schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/FailoverUrlSelector.java
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Returns true if the given Exception indicates the respective URL can be treated as failed.
+     *
+     * @param ex
+     */
+    protected boolean failedError(Exception ex) {
+        return true;
+    }
+
+    static class Entry implements Comparable<Entry> {
```
harshach: can we use some other name than Entry as its already used in Map.

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/LoadBalancedFailoverUrlSelector.java
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * This class defines a strategy to roundrobin among active URLs. When all URLs are considered failed which are maintained in
+ * the cache then it invalidates the cache and all urls are considered active.
+ */
+public class LoadBalancedFailoverUrlSelector extends AbstractUrlSelector {
+
+    /**
+     * Property to configure  time interval at which failed URL may be considered active.
+     */
+    public static final String FAILED_URL_EXPIRY_INTERVAL_MS = "failed.url.expiry.interval.ms";
```
harshach: probably its good time to place these configs in a single file.

harshach:
+1

PR #35, opened 2017-02-08
ISSUE#34 Support for optional schema evolution.

satishd:


harshach:
+1. @shahsank3t  we need to add a field in the UI that captures if the users want the schema to be evolved or not. Default should point to true and if user set this to false we should only allow them register one version of Schema.

satishd:
@shahsank3t @harshach Raised [JIRA](https://hwxiot.atlassian.net/browse/REGISTRY-23) for the earlier comment.

PR #39, opened 2017-02-21
Replaced gulp build process with webpack and added uglification step …

shahsank3t:
…after build


shahsank3t:
Added license and other UI lint rules such as proper indentation, trailing spaces, missing semi-colon, extra comma (similar to what we have in streamline).

cc: @harshach

PR #47, opened 2017-03-07
ISSUE#42 Added API to accept schema version as file instead of text on REST service and client

satishd:


satishd:
@shahsank3t You can use the newly added API for uploading schema versions.

PR #52, opened 2017-03-14
REGISTRY-14: Add a new module for spnego kerberos authorization

priyank5485:
I have disabled 3 tests related to kerberos since they require kdc setup. Initial request or any other request made by the browser to ui server works fine. However, there is a need to change some ui code so that subsequent http requests made by ui to the server on user clicking different controls also passes the same cookie that was generated as result of authentication. Currently, subsequent requests ignore the cookie and hence fail to authenticate. It should be a quick fix in the UI.


harshach:
@priyank5485 Nice Work!!. We also need to handle this for SchemaRegistryClient as it uses REST API to query .

harshach:
overall looks good to me. It would good to have SECURITY.md talking about the necessary configs to setup the secure env.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/KerberosAuthenticator.java
```diff
@@ -0,0 +1,370 @@
...
+ * The {@link KerberosAuthenticator} implements the Kerberos SPNEGO authentication sequence.
+ * <p>
+ * It uses the default principal for the Kerberos cache (normally set via kinit).
+ * <p>
+ * It falls back to the {@link PseudoAuthenticator} if the HTTP endpoint does not trigger an SPNEGO authentication
+ * sequence.
+ */
+public class KerberosAuthenticator implements Authenticator {
+
+    private static Logger LOG = LoggerFactory.getLogger(
```
satishd: nit: no new line required here.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/KerberosAuthenticator.java
```diff
@@ -0,0 +1,370 @@
...
+                @Override
+                public Void run() throws Exception {
+                    GSSContext gssContext = null;
+                    try {
+                        GSSManager gssManager = GSSManager.getInstance();
+                        String servicePrincipal = KerberosUtil.getServicePrincipal("HTTP",
+                                KerberosAuthenticator.this.url.getHost());
+                        Oid oid = KerberosUtil.getOidInstance("NT_GSS_KRB5_PRINCIPAL");
+                        GSSName serviceName = gssManager.createName(servicePrincipal,
+                                oid);
```
satishd: minor nit: avoid new line here.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/KerberosAuthenticator.java
```diff
@@ -0,0 +1,370 @@
...
+                        oid = KerberosUtil.getOidInstance("GSS_KRB5_MECH_OID");
+                        gssContext = gssManager.createContext(serviceName, oid, null,
+                                GSSContext.DEFAULT_LIFETIME);
```
satishd: minor nit: avoid new line here.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/KerberosAuthenticator.java
```diff
@@ -0,0 +1,370 @@
...
+                            if (!gssContext.isEstablished()) {
+                                inToken = readToken();
+                            } else {
+                                established = true;
+                            }
+                        }
+                    } finally {
+                        if (gssContext != null) {
+                            gssContext.dispose();
+                            gssContext = null;
```
satishd: nit: no need to set this as null as it is created in this method.

satishd:
@priyank5485 It seems large part of the code is borrowed from `hadoop-auth` module. Why can not we use that module and extend wherever it is possible instead of borrowing all those sources into this repo? 

cc @harshach

HeartSaVioR:
AFAIK, if the code is borrowed from hadoop-auth, the original license (copyright) should be kept.
That's what Heron does for borrowing Apache Storm's codebase.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/AuthenticatedURL.java
```diff
@@ -0,0 +1,278 @@
...
+     * Name of the HTTP cookie used for the authentication token between the client and the server.
+     */
+    public static final String AUTH_COOKIE = "hadoop.auth";
+
+    private static final String AUTH_COOKIE_EQ = AUTH_COOKIE + "=";
+
+    /**
+     * Client side authentication token.
+     */
+    public static class Token {
```
priyank5485: @satishd I dont see a strong reason for using the Token inner class. Its mainly to avoid returning new strings(tokens) as part of some method calls in client for a connection made to an already authenticated url. As you mentioned the code has been borrowed from hadoop-auth module and thats how it was implemented.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/AuthenticationException.java
```diff
@@ -0,0 +1,52 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+package com.hortonworks.registries.auth.client;
+
+/**
+ * Exception thrown when an authentication error occurrs.
```
priyank5485: Will fix.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/KerberosAuthenticator.java
priyank5485: Will fix.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/KerberosAuthenticator.java
priyank5485: Will fix.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/KerberosAuthenticator.java
priyank5485: Will fix.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/KerberosAuthenticator.java
priyank5485: Will fix.

priyank5485:
@harshach This module also has classes for non browser clients. I think I can incorporate them with schema registry client. Will update the PR. I will also add SECURITY.md file with it.

@satishd Main reason to create a module of our own is to not bring in some other dependencies like zookeeper or curator and get rid of code/classes we don't need. Even though classes can be extended and modified to change behavior to suit our needs, this gives us more control regarding changing code however we want to.

@HeartSaVioR Yes, all this code is borrowed from hadoop-auth module and changed slightly based on our needs. I have removed the Copyright Hortonworks part from the license which makes the license pretty much how it was on the original hadoop-auth files. Please let me know if this looks good or not

harshach:
@satishd @priyank5485 @HeartSaVioR we did the same thing in storm-core by making a dependency on hadoop-auth. It's turning into a PITA as we have topologies who depend on their own version of hadoop which also brings a different version of hadoop-auth. Yahoo! guys ran into issues they also tried removed this dependency from storm-core. I want to avoid potential issues because of that. It also makes it easy to change and improve as per our needs.

harshach:
Licensing info looks good to me.

HeartSaVioR:
@priyank5485 
https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-auth/src/main/java/org/apache/hadoop/util/PlatformName.java

Origin file clearly states that license is belong to ASF, not only describing that that is ASL v2.0. Let's keep the **origin license comment** and don't modify that.

satishd:
@priyank5485 @harshach @HeartSaVioR It is OK to borrow if the dependency issues are getting complicated with user code etc. But we will also be missing any fixes that go into `hadoop-auth` module which we have to keep an eye on that module before every streamline release and update these sources with `apache/hadoop-auth` fixes. 

 It may be good to document the reason of borrowing in README.md of this module.

harshach:
@satishd Agree on documenting the details and definitely we should keep track of changes going into hadoop-auth's SPNEGO.

hmcl:
@harshach @priyank5485 @HeartSaVioR @satishd I read your discussion about the `hadoop-auth` dependencies. There is no easy answer here attending to the whole dependencies discussion, but duplicating the source code will make it very difficult to keep track of any fixes that go into `hadoop-auth` on a timely manner. 

We also need similar code in Streamline. Should we duplicate the code in Streamline as well ?

I would suggest that If depending on `hadoop-auth` is really an impossibility, perhaps we should have a streamline/registry maven repo just with this code that we could maintain, and simply add a dependency in maven on both streamline and registry.

satishd:
@hmcl I guess the plan is to use this module in `streamline` by having a dependency on this auth module. @harshach may want to comment on that.

@priyank5485 new module for auth should be `common-auth` or `registry-auth` or some other relevant name instead of `schema-registry-auth` as this module is more of generic authentication and not really only about schema-registry. Schema registry related stuff can be pushed into a submodule in schema-registry.

schema-registry-auth/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>registries</artifactId>
+        <groupId>com.hortonworks.registries</groupId>
+        <version>0.1.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>schema-registry-auth</artifactId>
```
satishd: @priyank5485 You may want to rename this module as commented [here](https://github.com/hortonworks/registry/pull/52#issuecomment-288287062)

HeartSaVioR:
@hmcl 
I guess you're suggesting having common project. That would make sense to me, since Registry is based on Streamline, and shares lots of codebase.

We're taking two different approach simultaneously - copy and modify codes to make in sync, and referring other project's module - and neither is ideal for me. IMHO, Streamline just needs to refer Registry for provided client, not refer common module. (it only refers `Schema` class indeed...) If we have versions for two projects, it will be more complicated.

What @hmcl suggests are promising, but I'm OK to duplicate the codebase if it feels too ideal for us.

cc. @harshach @priyank5485 @satishd

satishd:
@HeartSaVioR registry is not based or dependent on streamline but it is the other way. I have put my opinion in my earlier [comment](https://github.com/hortonworks/registry/pull/52#issuecomment-288287062)

satishd:
@HeartSaVioR Let me elaborate on my earlier comment here. 
What I mean here is registry does not really depend upon any streamline module. We wanted storage component to be in registry as it is kind of individual component shared by both streamline and registry. 

It is ideal to have one more project which contains common stuff but it adds more logistics. It may be practical to have the common modules in streamline or registry and depend on respective modules.

cc @harshach @priyank5485 @hmcl

hmcl:
@HeartSaVioR @satishd @priyank5485 @harshach in essence all I mean is that many projects need an authentication (and security - for that matter) module just like streamline and registry now do. Currently Hadoop has that in it's own **hadoop-auth** module. The natural thing for me would be to have a project called **cluster-auth** or **cluster-security** or something along those lines, and refer to this project using maven coordinates. This project would have what currently exists in **hadoop-auth**.  In the limit, even Hadoop itself could depend on this project in the future (perhaps this is a bit too ambitious, but several projects started as a component of a project and became their own entity). 

This would be the ideal world scenario for me. In the short time frame maybe this is not doable. However, we can give some strides in this direction by putting all the necessary code in a security module in registry (since streamline depends on it), and have streamline refer to it. Perhaps in the future we can extract it to become it's own project. In essence, make it modular, and if we have to duplicate it for now, let's put it all in one place, and such a way that we can extract it at any point in the future.

For now I am OK with having this in registry and I will refer to it in streamline if referring to hadoop-auth is really really unpractical.

harshach:
@satishd @priyank5485 @hmcl @HeartSaVioR 
Agree with most points made here. But SPNEGO auth is only useful for UI based auth and it doesn't solve more complext Kerberos auth for services etc.. . Hence our initial idea was to port this into registry and build that as a independent module and use that in streamline.
I think points made here are aligns with this and Instead of making this as a independent project, for now lets make this as a module under registry. Streamline already depends on registry modules and we can use this module in Streamline.

@priyank5485 lets see if can add another module to registry under com.hortonworks.spengo-auth and not under registries. So that we can keep it as independent as it can be.
We are already publishing registry artifacts into maven central we will do the same with spengo-auth.

harshach:
@priyank5485 @satishd @HeartSaVioR @hmcl 

Just to make it clear the intentions here.

1. Registry is an independent project and doesn't need any dependencies from streamline.
2. Streamline depends on Registry for schemas and other things.
3. Registry & Streamline shares same storage & common part.

What are we trying to do:
1. Remove common code from Streamline repository
2. Move any modules that are appropriate to be in Registry like MLRegistry .
3. Make Streamline depend on registry storage & common modules
4. Use hadoop-auth SPNEGO code in registry module with modifications. (Reasons are already provided in above comments).
5. Make streamline depend on registry authentication module.

I hope the above clears up any confusion. There isn't any code duplication going on instead we are removing the current code duplication.

I hear your concerns about copying hadoop-auth code but it provides us felxibility of fixing/improving spengo auth instead of trying to patch hadoop-auth which involves a bigger community and lot less inclination to change code.

https://github.com/apache/hadoop/tree/trunk/hadoop-common-project/hadoop-auth/src/main/java/org/apache/hadoop/security/authentication/server

If you see the hadoop-auth module, there isn't whole lot of activity going on or there are frequent changes in the project. So we will be able to monitor and pick changes that are needed. This shouldn't be a huge concern at this point and current module works for us for the most part.

satishd:
@harshach default authorizer implementation([PR-517](https://github.com/hortonworks/streamline/pull/517)) is done in streamline which we may want to have it in registry because of the reasons mentioned in your last [comment](https://github.com/hortonworks/registry/pull/52#issuecomment-288479317). So, we may also need common-authorization module to be pushed to registry like other common modules.

priyank5485:
@satishd @hmcl @harshach @HeartSaVioR 

I went through the discussions above and it seems the consensus and the takeaway is to rename this module as per Satish's suggestion and putting it under a different group id as per Harsha's suggestion. We will need to revisit the authorizer code and probably moving it as well as suggested by Satish. If i am missing anything else here please let me know

harshach:
@satishd unlike authentication, authorizer will be a specific implementation as you are authorizing on resources in that project. I haven't gone through in details with Arun's implementation if thats a generic implementation we can definitely port it.

satishd:
@arunmahadevan Basic authorizer framework is independent of the resources which authorization can be applied. That can be extracted and reused in both streamline and registry. I am not advocating these modules should be in registry but it is good to share common modules. Discussed in earlier comments on how we would like to keep those common modules for now.
What do you think?

arunmahadevan:
@satishd the authorizer framework itself can be independent and re-used in registry, however a few things needs to be fixed and cleaned up.

1. Naming `StreamlineAuthorizer` -> `Authorizer`, mvn module `streamline-authorizer` -> `authorizer`
2. Some user facing interfaces are added in streamline-sdk. If we want it to be re-used, it should be within the authorizer module itself.
3. The underly tables are added to the streamline create and delete sql scripts. Maybe this needs to be separated out.

My preference is to do the above clean up and keep it in streamline for now and have it well tested. Whenever registry modules wants to use it, just include the maven dependency and fix any pending stuff to make it really independent. Later we can move it wherever we want. 

Overall I think we keep the modules wherever its most used/has the most dependency currently, and try to make it independent as and when required but refer in other project via maven dependencies and once it matures, we move it to the right place (if required)

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/AuthenticatedURL.java
```diff
@@ -0,0 +1,278 @@
+/**
+ * Copyright 2016 Hortonworks.
```
hmcl: @priyank5485 This one didn't get removed

webservice/src/main/java/com/hortonworks/registries/webservice/RegistryApplication.java
```diff
@@ -69,6 +72,21 @@ public void run(RegistryConfiguration registryConfiguration, Environment environ
         if (registryConfiguration.isEnableCors()) {
             enableCORS(environment);
         }
+
+        List<ServletFilterConfiguration> servletFilterConfigurations = registryConfiguration.getServletFilters();
```
hmcl: @priyank5485 let's put the code in lines 76-89 in their own method and call it something along the lines **addServletFilters** or **registerServletFilters**, following the pattern of small cohesive methods already existing in this **run** method. It will make the code much cleaner, modular, and easier to read.

schema-registry-auth/src/main/java/com/hortonworks/registries/auth/client/AuthenticatedURL.java
priyank5485: @hmcl I have not pushed that fix yet. Its on my local branch. I will in a bit

webservice/src/main/java/com/hortonworks/registries/webservice/RegistryApplication.java
priyank5485: @hmcl Will do

common-auth/SECURITY.md
```diff
@@ -0,0 +1,65 @@
+# Running registry and streamline web-services securely
+
+This module is intended to be used by registry and streamline web-services so that they can enable http client authentication via SPNEGO. Currently supported
```
priyank5485: @hmcl I did not get the part about any registry dependent web-services. Should we really add that?

common-auth/SECURITY.md
```diff
@@ -0,0 +1,65 @@
+# Running registry and streamline web-services securely
+
+This module is intended to be used by registry and streamline web-services so that they can enable http client authentication via SPNEGO. Currently supported 
+authentication mechanism is Kerberos. The code for this module has been borrowed from hadoop-auth(2.7.3) module in hadoop project and changed a little. The 
+reasons for doing so are to avoid having a dependency on hadoop-auth module which brings in some other modules, avoid conflicts with other versions of 
+hadoop-auth module and having more control over the changes needed in future. Some text for this document has been borrowed from SECURITY.md of Apache Storm
```
priyank5485: I don't mind removing that but it loses a lot of information that was suggested by others to be included in this doc. Lets keep it if you don't mind

common-auth/SECURITY.md
```diff
@@ -0,0 +1,65 @@
...
+
+This module is intended to be used by registry and streamline web-services so that they can enable http client authentication via SPNEGO. Currently supported 
+authentication mechanism is Kerberos. The code for this module has been borrowed from hadoop-auth(2.7.3) module in hadoop project and changed a little. The 
+reasons for doing so are to avoid having a dependency on hadoop-auth module which brings in some other modules, avoid conflicts with other versions of 
+hadoop-auth module and having more control over the changes needed in future. Some text for this document has been borrowed from SECURITY.md of Apache Storm  
+
+By default, registry and streamline web-services are running without authentication enabled and anyone can access the web-services from ui/client as far as they
+know the url and can access the web-server from the client machine. To enable authentication with the client, webservice needs to add a servlet filter from 
+this module. The webservice module will need to declare a dependency on this module. One way of adding a servlet filter in code is as follows. 
+
```
priyank5485: I am not sure if i understand correctly what the above means and if we should add that. May be i am missing something here.

common-auth/SECURITY.md
```diff
@@ -0,0 +1,65 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+}
+```
+
+ServletFilterConfiguration is a java object representing any servlet filter used in schema registry. However the general idea is that one needs to
+add com.hortonworks.registries.auth.server.AuthenticationFilter for enabling authentication 
+
+The configuration for that filter in yaml is represented in params property as follows.
```
priyank5485: I think its better to keep it. To give the reader a gist of what params is

common-auth/SECURITY.md
```diff
@@ -0,0 +1,65 @@
...
+     kerberos.keytab: "/path/to/keytab"
+     kerberos.name.rules: "RULE:[2:$1@$0]([jt]t@.*EXAMPLE.COM)s/.*/$MAPRED_USER/ RULE:[2:$1@$0]([nd]n@.*EXAMPLE.COM)s/.*/$HDFS_USER/DEFAULT"
+     token.validity: 36000
+```
+
+The servlet filter uses the principal `HTTP/{hostname}` (here hostname should be the host where the web-service runs) to login. Make sure that principal is 
+created as part of Kerberos setup
+
+Once configured, you must do `kinit` on client side using the principal of the user before accessing the web-service via the browser or some other client. 
+This principal also needs to be created first during Kerberos setup
```
priyank5485: thats for HTTP principal. This is for user principal. We should keep this

common-auth/SECURITY.md
```diff
@@ -0,0 +1,65 @@
...
+
+Here's an example of accessing web-service after the setup above:
+```bash
+curl  -i --negotiate -u:anyUser  -b ~/cookiejar.txt -c ~/cookiejar.txt  http://<web-service-host>:<port>/api/v1/
```
priyank5485: cookie jar is the file where cookies are stored when server first respnds with the cookie and we are passing the same back to the server in the request. I think this is more for man curl. We can avoid explaing how curl works here

common-auth/SECURITY.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+1. Firefox: Go to `about:config` and search for `network.negotiate-auth.trusted-uris` double-click to add value "http://<web-service-host>:<port>"
+2. Google-chrome: start from command line with: 
+   `google-chrome --auth-server-whitelist="*web-service-hostname" --auth-negotiate-delegate-whitelist="*web-service-hostname"`
+3. IE: Configure trusted websites to include "web-service-hostname" and allow negotiation for that website
+
+**Caution**: In AD MIT Kerberos setup, the key size is bigger than the default UI jetty server request header size. If using that, make sure you set 
+header buffer bytes to 65536
```
priyank5485: In the request header

priyank5485:
@hmcl I have addressed most of your review comments on SECURITY.md. Put questions or replies where i was not sure. Definitely agree on common base module.

common-auth/SECURITY.md
hmcl: the text implies that the header buffer is to be set in the jetty server, which according to what you say is not the case. The text should be clear

priyank5485: What is the text you prefer? I will update it to that

common-auth/SECURITY.md
hmcl: Well, it says the same exact thing succinctly.

common-auth/SECURITY.md
hmcl: This paragraph is very confusing to me. My understanding is: if a web service wants to use authentication, and that web service declares registry as a dependency, then the code snippet below illustrates how to do create a servlet filter. That's what I meant to say with my suggestion.If you want you can replace dependent with declares registry as dependency, or you can amend your original paragraph to be more clear.

common-auth/SECURITY.md
hmcl: oh, I know now what you mean. I would write it as:

The filter configuration is passed using the params property in they YAML file, as follows:

common-auth/SECURITY.md
hmcl: OK, but since I didn't understand that when I read it, perhaps we should try to make the text a bit more clear.

common-auth/SECURITY.md
hmcl: ok, but if it was confusing to me, won't it be confusing to users reading this doc ?

common-auth/SECURITY.md
hmcl: **Caution**: In AD MIT Kerberos setup, the key size is bigger than the default UI jetty server request header size. If using MIT Kerberos, make sure you set the **(HTTP ?)** request header buffer bytes to 65536.

common-auth/SECURITY.md
priyank5485: This paragraph is just saying that if a webservice needs to use authentication they can use this module. It does not need to declare any dependency on registry top level maven module. Just the auth module here. The code snippet illustrates how you can add a add a servlet filter(which is why it says One way of adding.....) and shows how its done in registry web-service project. Code to add servlet filter is not part of this module. And they can add servlet filters in other ways too. Is it still unclear? Do you still want any modification to the paragraph?

common-auth/SECURITY.md
priyank5485: Will change

common-auth/SECURITY.md
priyank5485: Any suggestions for text?

common-auth/SECURITY.md
priyank5485: It could be. How do you want me to word it? We described above what needs to be done on web server side and client side from Kerberos point of view and then we are just giving an example of a client aka curl. Even the settings change in browser is not clear and confusing. I did not understand when i read it in the first place. I think its out of the scope trying to explain documentation about browser settings or curl

common-auth/SECURITY.md
priyank5485: Will change.

priyank5485:
@hmcl pushed one more fix with a couple of changes

harshach:
+1

PR #62, opened 2017-04-17
REGISTRY-61: Add a convenient way to pass JAAS config file.

harshach:


harshach:
@priyank5485 can you take a look.

PR #77, opened 2017-04-29
ISSUE-74: Added API to throw an error if a schemametadata is added violating unique constraints

satishd:


satishd:
You can add `_throwErrorIfExists=true` request parameter to `POST /schemas` to receive any errors while adding the schema metadata.

shahsank3t:
@satishd I am making a POST call via API: "api/v1/schemaregistry/schemas?_throwErrorIfExists=true" and the payload has the same (schema) name which is already present in the registry. I am still not getting any exception in the response but the previous schemaId.

Can you please verify the same from your end?

satishd:
Verified backend registry and dao API works with tests. But HTTP POST from webservice does not work as POST request does not really pass request parameters. This is because POST request is sent with json payload and query parameters are ignored by the client/UI while sending the HTTP request. 
POST request can also be sent with other content-types like `application/x-www-form-urlencoded` or `multipart/form-data` but it creates a backward incompatible API change. One possible hacky solution is to receive the same header as HTTP header. This API is used only in UI for now.

shahsank3t:
@satishd To avoid any kind of hack, why not go for query parameter instead of request parameter? Just like we have for "_orderByFields". What do you say?

satishd:
@shahsank3t Explained in earlier comment, meant request parameters as request query parameters.

satishd:
`_orderByFields` is used in GET request but not in POST request, that is why it works fine.

satishd:
@shahsank3t Can you try this patch with respective UI changes?

shahsank3t:
+1 for me, thanks @satishd for adding the UI change as well into the PR.

PR #97, opened 2017-05-09
ISSUE-96: Resolved postgres dependencies

satishd:


pom.xml
```diff
@@ -462,6 +463,12 @@
                 <artifactId>spring-context</artifactId>
                 <version>${spring.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.postgresql</groupId>
+                <artifactId>postgresql</artifactId>
+                <version>${postgresql.version}</version>
+                <scope>provided</scope>
```
harshach: we need to copy this into the libs via dist build. Since this is BSD-2 we should be ok with ASLv2 license

satishd:
@harshach Pushed the change to have postgres lib as part of dependency and it will be copied to libs with the distribution.

harshach:
+1

PR #101, opened 2017-05-10
ISSUE-99: Kafka avro serializer/deserializer should treat schema name as topic name by default

satishd:


harshach:
@satishd do we throw a error if the user registered lets say "truck_events" and now tries to register a "truck_events:v" , vice versa

satishd:
@harshach We do not throw an error in those scenarios.

satishd:
@harshach No error is thrown, default implementation maps schema name as topic name for values. This can be over ridden by user extending `KafkaAvroSerializer`.  This is to keep this contract simple. Respective changes in streamline are done once this PR is merged.

harshach:
+1

PR #106, opened 2017-05-11
ISSUE-102: Simplify APIs for registering/fetching serdes api.

satishd:
- upload ser/des jar 
- register serdes info 
- map serdes with a schema

Detailed documentation with json structure can be seen with swagger.


harshach:
+1

PR #111, opened 2017-05-16
ISSUE-107: Generic search API implementation in storagemanager supporting SQL like constructs.

### LINKED ISSUE #107: Search API in finding schemas with name, description.
Currently, there is search API to find schema versions containing fields with a given name. There should be a search API to find schemas for given name or description.

satishd:
- Generic search API implementation in storagemanager supporting SQL like constructs.
- Added search for schemas with name and description

Closes #107


satishd:
UI can consume this PR for providing search based on name and description like below. But backend storagemanager API is powerful to have SQL like queries with where and orderby clauses. 
More details can be found at respective swagger API.

```
curl -X GET --header 'Accept: application/json' 'http://localhost:9090/api/v1/schemaregistry/schemas/search'
```

satishd:
@shahsank3t You can use this API with postgres/mysql and let me know if you need more info.

shahsank3t:
@satishd this API will send back the response based on "schemaregistry/schemas" endpoint but now UI is using "schemaregistry/schemas/aggregated" endpoint. Can you make the appropriate changes for it and also rebase it. Thanks.

satishd:
@shahsank3t Added api for getting aggregated schema for given search.

```curl -X GET --header 'Accept: application/json' 'http://localhost:9090/api/v1/schemaregistry/search/schemas/aggregated?name=<>&description=<>'```

**Comment by shahsank3t** [comment]
Thanks @satishd 
Will this API also support sorting? If a user wants to search for schema's and then wants to sort them, we should have that flexibility.

**Comment by satishd** [comment]
Backend api supports sorting but we need to find a clean way to expose through REST.

**Comment by satishd** [comment]
Current REST api supports name/description fields but generic REST api can be added. Lets not rush through now for having right API, we can add later.

**Comment by satishd** [comment]
@shahsank3t Added orderBy also and order by query parameter can be like below.
```
_orderByFields=[<field-name>,<a/d>,]*
For ex: parameter can be  "_orderByFields=foo,a,bar,d" order by foo with ascending then bar with descending
```
URL can be like 
```
http://localhost:9090/api/v1/schemaregistry/search/schemas/aggregated?name=<>&description=<>&_orderByFields=name,a
```

satishd:
We can standardize REST based search query later.

shahsank3t:
+1
Thanks for the changes @satishd

Issue #99, opened 2017-05-10
Kafka avro serializer/deserializer should treat schema name as topic name by default.


satishd:
This issue is raised based on the discussion happened on google-groups [conversation](https://groups.google.com/forum/#!topic/registry/2HcXO3j6Iws).

By default Kafka Avro serializer computes schema name as the topic name. User can always over ride this default behavior by implementing [getSchemaKey](https://github.com/hortonworks/registry/blob/master/schema-registry/serdes/src/main/java/com/hortonworks/registries/schemaregistry/serdes/avro/kafka/KafkaAvroSerializer.java#L62) in [KafkaAvroSerializer](https://github.com/hortonworks/registry/blob/master/schema-registry/serdes/src/main/java/com/hortonworks/registries/schemaregistry/serdes/avro/kafka/KafkaAvroSerializer.java)

satishd:
PR #101 resolves this issue.

PR #123, opened 2017-05-22
ISSUE-132 Allow customising key/value suffix by configuration

michaelandrepearce:
Improve on https://github.com/hortonworks/registry/issues/99
Make the suffix configurable, as such reducing the need for custom concrete extensions of the serializer.


satishd:
@michaelandrepearce Do you use prefix/suffix for schema names in your scenarios?

satishd:
Closes #132

satishd:
@michaelandrepearce We are following PR description and commits to start with respective issue  like "ISSUE-XXX <message>"
I will take care of the current PRs and commits for now, you can follow this convention for next PRs.

michaelandrepearce:
@satishd sure. re use case, it was actually inline with the general work I'm trying to make migration / integration from confluent easier. In confluent they use -key and -value suffix's. 

So this paired with the confluent api, and then the protocol changes would mean you could inter-op where people have confluent serdes without changing code, just can redirect the schema registry url.

PR #144, opened 2017-06-07
ISSUE-136: Handle default and aliases attributes in schema normalization

### LINKED ISSUE #136: Not able to add default value in schema once the schema is saved.
1. Create a schema with some fields E.g.
```
{
 "type": "record",
 "name": "test",
 "namespace": "com.hortonworks.registries",
 "fields": [
  {
   "name": "business_id",
   "type": "string"
  }
 ]
}
```

2. Save

3. Click edit and now change the definition of field as 

```
  {
   "name": "business_id",
   "type": "string",
    "default": "x"
  }
```

4. Enter some description
5. Save

UI throws a message saying "Schema version already present" and does not save the changes.

satishd:
Closes #136

### DIFF
--- a/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/avro/AvroSchemaProvider.java
+++ b/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/avro/AvroSchemaProvider.java
@@ -15,6 +15,7 @@
  **/
 package com.hortonworks.registries.schemaregistry.avro;
 
+import com.google.common.base.Joiner;
 import com.hortonworks.registries.schemaregistry.AbstractSchemaProvider;
 import com.hortonworks.registries.schemaregistry.CompatibilityResult;
 import com.hortonworks.registries.schemaregistry.SchemaCompatibility;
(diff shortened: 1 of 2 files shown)


arunmahadevan:
From [here](https://issues.apache.org/jira/browse/AVRO-2002), 

"Identical fingerprints mean that one schema can read output of the other without schema resolution"

In that case would it make sense to just update the matching schema version after checking compatibility with any previous versions rather than changing the fingerprint logic and adding a new schema. Need to think through the different cases, but seems it may work. This will also handle the cases where users update the doc or other non relevant fields.

satishd:
IMO, it may be better to add a new version for default and aliases to keep this simple. We can have update API for each version to update attributes like doc etc. Raised issue #145 for that.

satishd:
Whenever there is an update to a specific version of schema which changes the semantics then it will be added as a new version. Attributes like default and aliases have an affect. These schemas can be used as projection schemas by consumers and that version should not be modified with default for any field as it would effect the behavior.

arunmahadevan:
So If I have a "backward" compatible schema with below fields,

```
{
  ...
 "fields": [  
  {"name": "A", "type": "int","default": 0},
  {"name": "B", "type": "int"}
  ]
}
```

And I edit and remove `"default":0` from "A", it gets saved as a new version.

However the data that the "V1" could project / deserialize (e.g. {"B": 100}) cannot be handled by "V2" version of the schema. Maybe more compatibility checks are needed before you create the new version.

arunmahadevan:
If the data thats going to be de-serialized is always written via the same SR instance, the incompatibility doesn't arise. i.e. for writing `{"B":100}` there needs to be a schema with `"fields": [{"name":"B", "type": "int"}]` which will be incompatible if one tries to update the schema with `"fields": [ {"name": "A", "type": "int"}, {"name":B", "type":"int"}]`

arunmahadevan:
+1

Issue #131, opened 2017-05-25
Make wire protocol more efficient.


michaelandrepearce:
The selling point of using a schema registry is to reduce the bytes in flight, so that we can send data as efficient as possible as in large mass data throughput system every byte really counts, also to avoid the schema or meta data dwarfing the real data we want to send.

Currently wire protocol uses up 12 bytes for just identifying the schema of the data, when in fact in we have a unique 8 byte (int64 - java long) id available to us (SchemaVersionStorable id).

This could even be reduced just to 4 byte (int32 - java int) to reduce further by using one protocol id byte to send if the id < Int.Max (and just using 4 bytes) and another protocol id to send if the id > Int.Max (and then using 8 bytes)

Current: 
(protocolId = 0x1)
```
        byteArrayOutputStream.write(new byte[]{0x1});

        // 8 bytes : schema metadata Id
        // 4 bytes : schema version
        byteArrayOutputStream.write(ByteBuffer.allocate(12)
                                            .putLong(schemaIdVersion.getSchemaMetadataId())
                                            .putInt(schemaIdVersion.getVersion()).array());
```
Proposed:
(protocolId = 0x1, or 0x2)
```
if (SchemaVersionStorable.id < Int.max) {
        byteArrayOutputStream.write(new byte[]{0x1});

        // 4 bytes : schema version id (as int)
        byteArrayOutputStream.write(ByteBuffer.allocate(4)                
                                            .putInt( (int) SchemaVersionStorable.id).array());
} else {
        byteArrayOutputStream.write(new byte[]{0x2});

        // 8 bytes : schema version id (as long)
        byteArrayOutputStream.write(ByteBuffer.allocate(8)                
                                            .putLong( SchemaVersionStorable.id).array());
}
```
Saving 8 bytes!!!! when the id < Max.Int which in reality is unlikely to reach this man, but still we cater for incase.

satishd:
@michaelandrepearce Thanks for bringing this issue. It was discussed earlier among ourselves before making this repo public and there was an internal JIRA filed. 

Custom protocol support is added to allow users have their own protocol and they do not necessarily use the default when it does not suit their needs.

Long can be avoided as it does not make much sense in reality for this usecase. Integer gives ~4.2billion numbers and 2,147,483,648 positive ids which means 21+ million schemas with 100 versions, very unlikely to reach this limit. We can even have logic to wrap to negative ids but that may be unnecessary.

michaelandrepearce:
[...]

Issue #126, opened 2017-05-23
Add support of handling confluent kafka byte wire protocol


michaelandrepearce:
To aid migration for user currently using the confluent platform, where producers are using still confluent serdes
or 
Where need to integrate with tooling that supports the confluent serdes only atm (until registry adoption is more wide spread).

It would be good to be able to produce a confluent wire compatible protocol, and like wise consume, making the serialiser configurable which protocol to produce, and the consumer able to simply just handle either.

This will need to ensure we check the protocol versions in the byte array, confluent currently uses a leading 0x0 byte for the protocol magic byte, followed by 4 bytes for the int32 id.

This would link with the other confluent compatibility work I've already PR'd, and enable a faster adoption.

michaelandrepearce:
Note this will be dependant on https://github.com/hortonworks/registry/issues/120. So will need this merged first.

michaelandrepearce:
this is linked to: https://github.com/hortonworks/registry/issues/131

PR #150, opened 2017-06-20
ISSUE-131 Add serdes protocol APIs

satishd:
- Added serdes protocol APIs
- Added wire level protocol carrying schema version instance id
- Added tests


satishd:
Closes #131

satishd:
ping @michaelandrepearce

michaelandrepearce:
@satishd i went through this, it seems that the protocol id for the int key protocol is 0x2 i was expecting this to be 0x1 and the protocol to be in line with the confluent's one (e.g. special case of byte[] handled else expects avro schema (no text/string special case) like wise no handling/wrapping of primitives, so it is swappable. 

Did i miss something?

schema-registry/serdes/src/main/java/com/hortonworks/registries/schemaregistry/serdes/avro/AvroSnapshotSerializer.java
```diff
@@ -50,10 +51,24 @@ public AvroSnapshotSerializer(ISchemaRegistryClient schemaRegistryClient) {
...
     protected byte[] doSerialize(Object input, SchemaIdVersion schemaIdVersion) throws SerDesException {
         try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
-            writeProtocolId(CURRENT_PROTOCOL_VERSION, byteArrayOutputStream);
-            writeSchemaVersion(schemaIdVersion, byteArrayOutputStream);
+            SchemaVersionProtocolHandlerRegistry
+                    .get()
+                    .getSerDesProtocolHandler(protocolVersion)
+                    .handleSchemaVersionSerialization(byteArrayOutputStream, schemaIdVersion);
+
             writeContentPayload(input, byteArrayOutputStream);
```
michaelandrepearce: Should the way the content is serialised/derserialized not also go into the protocol handler, e.g. confluent case is the protocol is slightly different where Strings are not a special case, and like wise primitives are wrapped

schema-registry/serdes/src/main/java/com/hortonworks/registries/schemaregistry/serdes/avro/SchemaVersionProtocolHandlerRegistry.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Registered ser/des protocol handlers which can be used in serializers/deserializers.
+ */
+public final class SchemaVersionProtocolHandlerRegistry {
+    public static final byte METADATA_ID_VERSION_PROTOCOL = 0x1;
+    public static final byte VERSION_ID_AS_LONG_PROTOCOL = 0x2;
+    public static final byte VERSION_ID_AS_INT_PROTOCOL = 0x3;
```
michaelandrepearce: should this not be 0x0? As int protocol intent was to align with confluents protocol a little (two birds one stone)

satishd: I do not want to mix our protocol with confluent's. We can continue to have our protocol with version identifier as int and our serialized data format. That is why it is not added as part of protocol version 0x0, which should be according to Confluent's serialization.

schema-registry/serdes/src/main/java/com/hortonworks/registries/schemaregistry/serdes/avro/AvroSnapshotSerializer.java
satishd: I wanted to start with how versioned schema identifier are encoded as part of protocol definition. Added javadoc comment explaining this in `SchemaVersionProtocolHandler`. I am fine in adding the actual serialization of payload as part of protocol in the next step.

michaelandrepearce: cool, will have to do this for https://github.com/hortonworks/registry/issues/126

schema-registry/serdes/src/main/java/com/hortonworks/registries/schemaregistry/serdes/avro/SchemaVersionProtocolHandlerRegistry.java
michaelandrepearce: ok no worries, in that case as i can't can we make https://github.com/hortonworks/registry/issues/126 targeted for next release, and once this is merged ill have to do this asap.

satishd: @michaelandrepearce I am enhancing this to include payload serdes and start making appropriate changes so that it would be easy for you to push confluent related protocol as one more protocol. Need to make these changes by without breaking current serializer/deserializer contracts. I plan to finish this by Monday and you can add confluent protocol after that.

satishd:
@michaelandrepearce Added ser/des protocol support for payloads also. Once this is merged, you may want to add confluent ser/des protocol for your scenarios.

michaelandrepearce:
Cool looks like it should work nicely for me. Thanks :) +1

Issue #159, opened 2017-06-22
Fix the install document


vesense:
1. Rename the db name from `registry_db` to `schema_registry`, because the `registry.yaml`  uses `schema_registry` as the default db name. Or new users will get error.
2. The step "Run bootstrap scripts" for CentOS/RetHat is missing.
3. Also added "Download latest Schema Registry binaries" step for CentOS/RetHat.

PR #160, opened 2017-06-22
ISSUE-159: Fix the install document

vesense:
1. Rename the db name from `registry_db` to `schema_registry`, because the `registry.yaml`  uses `schema_registry` as the default db name. Or new users will get error.
2. The step "Run bootstrap scripts" for CentOS/RetHat is missing.
3. Also added "Download latest Schema Registry binaries" step for CentOS/RetHat.


harshach:
+1.

satishd:
Closes #159

satishd:
Thanks @vesense

Issue #130, opened 2017-05-24
ISchemaRegisty and ISchemaRegistyClient alignment.


michaelandrepearce:
Over time it seems these two interfaces have slowly mis-aligned them selves.

Ideally a client should have the same interface as its service. As discussed to align this without breaking client compatibility, it makes sense to make the service interface inline with the client.

After such work it would be good to make one interface extend the other, to reduce this mis-alignment in future.

PR #162, opened 2017-06-26
ISSUE-130 Align server/client side apis with service api.

satishd:
- Align server/client side apis with service api.    
- Aligned without breaking the existing client APIs.

### DIFF
--- a/schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/ISchemaRegistryClient.java
+++ b/schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/ISchemaRegistryClient.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2016 Hortonworks.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -16,24 +16,15 @@
 
 package com.hortonworks.registries.schemaregistry.client;
 
-import com.hortonworks.registries.schemaregistry.SchemaFieldQuery;
+import com.hortonworks.registries.schemaregistry.ISchemaRegistryService;
 import com.hortonworks.registries.schemaregistry.SchemaIdVersion;
-import com.hortonworks.registries.schemaregistry.SchemaMetadata;
-import com.hortonworks.registries.schemaregistry.SchemaMetadataInfo;
-import com.hortonworks.registries.schemaregistry.SchemaProviderInfo;
-import com.hortonworks.registries.schemaregistry.SchemaVersion;
-import com.hortonworks.registries.schemaregistry.SchemaVersionInfo;
-import com.hortonworks.registries.schemaregistry.SchemaVersionKey;
 import com.hortonworks.registries.schemaregistry.SerDesInfo;
-import com.hortonworks.registries.schemaregistry.SerDesPair;
 import com.hortonworks.registries.schemaregistry.errors.IncompatibleSchemaException;
 import com.hortonworks.registries.schemaregistry.errors.InvalidSchemaException;
 import com.hortonworks.registries.schemaregistry.errors.SchemaNotFoundException;
 import com.hortonworks.registries.schemaregistry.serde.SerDesException;
 
-import java.io.FileNotFoundException;
 import java.io.InputStream;
-import java.util.Collection;
 
 /**
  * This interface defines different methods to interact with remote schema registry.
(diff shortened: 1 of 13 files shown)


satishd:
Closes #130

harshach:
+1

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/ISchemaRegistryClient.java
```diff
@@ -97,176 +90,51 @@
...
-
-
     /**
      * @param schemaName   name identifying a schema
      * @param toSchemaText text representing the schema to be checked for compatibility
-     * @return true if the given {@code toSchemaText} is compatible with the latest version of the schema with id as {@code schemaName}.
+     * @return true if the given {@code toSchemaText} is compatible with all the versions of the schema with name as {@code schemaName}.
      * @throws SchemaNotFoundException if there is no schema metadata registered with the given {@code schemaName}
      */
     boolean isCompatibleWithAllVersions(String schemaName, String toSchemaText) throws SchemaNotFoundException;
```
michaelandrepearce: Should this not be declared at a service level, should be a remove call.

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ISchemaRegistryService.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    Collection<SchemaProviderInfo> getSupportedSchemaProviders();
+
+    /**
+     * Registers information about a schema.
+     *
+     * @param schemaMetadata metadata about schema.
+     * @return id of the registered schema which is successfully registered now or earlier.
+     */
+    Long registerSchemaMetadata(SchemaMetadata schemaMetadata);
```
michaelandrepearce: for the small breaking change should we not rename this as add, or rename the adds as register (seems inconsistent)

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/errors/UnsupportedSchemaTypeException.java
```diff
@@ -18,7 +18,9 @@
 /**
  * Indicates that schema type is invalid.
  */
-public class UnsupportedSchemaTypeException extends Exception {
+public class UnsupportedSchemaTypeException extends RuntimeException {
```
michaelandrepearce: Do we really need to be making RuntimeExceptions, its a lot harder for end users to handle as they're easily left off interfaces etc. We should be avoiding IMO runtime exceptions

michaelandrepearce:
@satishd thanks, have left some comments, cheers.

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/errors/UnsupportedSchemaTypeException.java
satishd: This Exception is not really exposed on client side. Extending RTE is a controversial topic and it has both advantages/disadvantages. My preference is to avid checked exceptions.

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/ISchemaRegistryClient.java
satishd: There is a different method signature in `ISchemaRegistry` and that is the reason why this was not moved to `ISchemaRegistryService`. Ideally, I would have deprecated this method and introduced the existing one in `ISchemaRegistry`. I was not sure whether I would add that now, let me know your opinion.

satishd:
@michaelandrepearce Thanks for reviewing, updated with my comments.

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/ISchemaRegistryClient.java
michaelandrepearce: I think if the effort is to try align then where we can align be deprecating and old and adding a new it's worth it, as long as there's a pre agreed time the deprecated would be removed eg would revomed in next major +1 release

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ISchemaRegistryService.java
michaelandrepearce: Fair enough. Maybe as per the other comment, add the aligned name and keep the deprecated one with agreed removal time scales

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/errors/UnsupportedSchemaTypeException.java
michaelandrepearce: Fair enough

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/ISchemaRegistryClient.java
satishd: @harshach I plan to change couple of these APIs as deprecated and remain till next release. Let me know your opinion.

harshach: @satishd  sounds good to me

satishd:
@michaelandrepearce @harshach Updated with deprecated API changes.

harshach:
still +1 on merging this.

PR #174, opened 2017-07-03
ISSUE-172 Support Disabling Schemas

michaelandrepearce:
Add ability to disable schema's this will mean when disabled, a schema will no longer partake in compatibilty checks (e.g. allowing disabling an accidental registration). 
Also address's ISSUE-173, by providing helper method ability to disable all versions of a given schema (for sakes of topic deletion)

### DIFF
--- a/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/SchemaVersionInfo.java
+++ b/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/SchemaVersionInfo.java
@@ -51,18 +51,24 @@ public final class SchemaVersionInfo implements Serializable {
      * timestamp of the schema which is given in SchemaInfo
      */
     private Long timestamp;
+
+    /**
+     * if the schema version is disabled which is given in SchemaInfo
+     */
+    private Boolean disabled;
     
 
     @SuppressWarnings("unused")
     private SchemaVersionInfo() { /* Private constructor for Jackson JSON mapping */ }
 
-    public SchemaVersionInfo(Long id, String name, Integer version, String schemaText, Long timestamp, String description) {
+    public SchemaVersionInfo(Long id, String name, Integer version, String schemaText, Long timestamp, Boolean disabled, String description) {
         this.id = id;
         this.name = name;
         this.description = description;
         this.version = version;
         this.schemaText = schemaText;
         this.timestamp = timestamp;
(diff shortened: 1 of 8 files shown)


harshach:
@michaelandrepearce Thanks for the patch. This is the most asked feature/request often. But we've a different take on it, to support Schema Lifecycle Management.
1. When user first registers a schema it goes into TEST state.
2. if this schema is tested and peer approved it can promoted to PRODUCTION state.
3. Clients by default can only see the schema in PRODUCTION state and can be configured to see TEST state schema as well. 
4. If there are no clients using the schemas (client heartbeat functionality will show who all are accessing a schema) than it can go into DISABLED state.

We should at least start recording "STATE" as another column and make this patch to record this into the said column. We can add the above on top of it.

michaelandrepearce:
these two things I think are independent. What you're talking about is environment promotion. Test to uat to prod, very much like an artefact repo.

 What we are needing is the ability to delete a schema (or at least make it ignored by back compatibility checks).

In your case where client no longer used if I had a valid schema even after client in prod stopped using it I would want the schema validated historically as it would still be possible to retro activate a client. Or at least if we are to merge the two I would suggest having a different state for this

michaelandrepearce:
the other issue slightly with hard coded environments or merging the two ideas and enum the env state, will be each org will have separate setup of environment promotion. Eg it's likely in a big org you can have Dev, test, qa, uat, demo, live. And no doubt other orgs all have different. Typically in a commercial artefact repo this is completely configurable for the staging layers. But globally the ability to disable/delete an artefact is a seperate function

michaelandrepearce:
For this reason I would suggest life cycle management and ability to disable/delete are left as two seperate features.

michaelandrepearce:
@satishd @harshach I assume by creation of https://github.com/hortonworks/registry/issues/177, we're agreeing to separate the concerns.

satishd:
@michaelandrepearce @harshach I did not mean that these two are separate, this issue may be one aspect of schema lifecycle management.  Lifecycle is not only about promotion but includes stages from creation to till it is archived/deleted. We need to think through in detail and give a way to plug any states in between by users. This does not need to cover all usecases but it should be pluggable enough with basic lifecycle management. I will put my thoughts in #177 and discuss in detail over there.

michaelandrepearce:
@satishd I'm closing this off.

Issue #171, opened 2017-07-03
Add Validate Latest option while adding new schema versions.


satishd:
This issue is raised to address [discussion](https://github.com/hortonworks/registry/issues/166#issuecomment-311861229) happened on #166.

This was needed for Confluent SR migration as it has an option only to validate against earlier schema. This should not be the default behavior and a good practice is to have the option set as validate against all earlier versions of the schema. 
Current behavior of validating against all earlier versions of the schema continues to be the default option.

michaelandrepearce:
@satishd should add the need/ability to change the compatibility setting with this

satishd:
@michaelandrepearce I started working on this yesterday evening but I got a notification later about  PR addressing this issue. Please send an email to registry-group or assign issue to yourself when you plan to work on any issues so that there will not be any duplicate work from others. I also should have done the same. :)

Thanks for raising the PR.

PR #175, opened 2017-07-03
ISSUE-171 Support Validation Level

michaelandrepearce:
Add ability to validate ALL or LATEST

### DIFF
--- a/schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/DefaultSchemaRegistry.java
+++ b/schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/DefaultSchemaRegistry.java
@@ -191,7 +191,7 @@ public class DefaultSchemaRegistry implements ISchemaRegistry {
             final Long nextId = storageManager.nextId(givenSchemaMetadataStorable.getNameSpace());
             givenSchemaMetadataStorable.setId(nextId);
             givenSchemaMetadataStorable.setTimestamp(System.currentTimeMillis());
-            storageManager.add(givenSchemaMetadataStorable);
+            storageManager.addOrUpdate(givenSchemaMetadataStorable);
             return givenSchemaMetadataStorable.getId();
         }
     }
@@ -291,35 +291,18 @@ public class DefaultSchemaRegistry implements ISchemaRegistry {
                 try {
                     Integer version = 0;
                     if (schemaMetadata.isEvolve()) {
-                        Collection<SchemaVersionInfo> schemaVersionInfos = getAllVersions(schemaName);
-                        if (schemaVersionInfos != null && !schemaVersionInfos.isEmpty()) {
-                            SchemaVersionInfo latestSchemaVersionInfo = null;
-                            SchemaCompatibility compatibility = schemaMetadata.getCompatibility();
-                            for (SchemaVersionInfo schemaVersionInfo : schemaVersionInfos) {
-                                // check for compatibility
-                                CompatibilityResult compatibilityResult =
-                                        schemaTypeWithProviders
-                                                .get(type)
-                                                .checkCompatibility(schemaText,
-                                                                    schemaVersionInfo.getSchemaText(),
-                                                                    compatibility);
-                                if (!compatibilityResult.isCompatible()) {
-                                    String errMsg = String.format("Given schema is not compatible with earlier schema versions. \n" +
-                                                                          "Error location: [%s] \n" +
-                                                                          "Error encountered is: [%s]",
-                                                                  compatibilityResult.getErrorLocation(),
-                                                                  compatibilityResult.getErrorMessage());
-                                    LOG.error(errMsg);
-                                    throw new IncompatibleSchemaException(errMsg);
-                                }
-
-                                Integer curVersion = schemaVersionInfo.getVersion();
-                                if (curVersion >= version) {
-                                    latestSchemaVersionInfo = schemaVersionInfo;
-                                    version = curVersion;
-                                }
-                            }
-
+                        CompatibilityResult compatibilityResult = checkCompatibility(schemaName, schemaText);
+                        if (!compatibilityResult.isCompatible()) {
+                            String errMsg = String.format("Given schema is not compatible with latest schema versions. \n" +
+                                                          "Error location: [%s] \n" +
+                                                          "Error encountered is: [%s]",
+                                                          compatibilityResult.getErrorLocation(),
+                                                          compatibilityResult.getErrorMessage());
+                            LOG.error(errMsg);
+                            throw new IncompatibleSchemaException(errMsg);
+                        }
+                        SchemaVersionInfo latestSchemaVersionInfo = getLatestSchemaVersionInfo(schemaName);
+                        if (latestSchemaVersionInfo != null) {
                             version = latestSchemaVersionInfo.getVersion();
                         }
                     }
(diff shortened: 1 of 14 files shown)


michaelandrepearce:
address's https://github.com/hortonworks/registry/issues/171

schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/DefaultSchemaRegistry.java
```diff
@@ -278,31 +278,52 @@ private SchemaVersionInfo createSchemaVersion(SchemaMetadata schemaMetadata,
                         if (schemaVersionInfos != null && !schemaVersionInfos.isEmpty()) {
                             SchemaVersionInfo latestSchemaVersionInfo = null;
                             SchemaCompatibility compatibility = schemaMetadata.getCompatibility();
+                            SchemaValidationLevel validation = schemaMetadata.getValidation();
                             for (SchemaVersionInfo schemaVersionInfo : schemaVersionInfos) {
+                                if (SchemaValidationLevel.ALL.equals(validation)) {
+                                    // check for compatibility
+                                    CompatibilityResult compatibilityResult =
```
satishd: Below code can be extracted into a method and which can also be used for condition `SchemaValidationLevel.LATEST.equals(validation)`

```
    // check for compatibility
    CompatibilityResult compatibilityResult =
        schemaTypeWithProviders
            .get(type)
            .checkCompatibility(schemaText,
                                schemaVersionInfo.getSchemaText(),
                                compatibility);
    if (!compatibilityResult.isCompatible()) {
        String errMsg = String.format("Given schema is not compatible with earlier schema versions. \n" +
                                      "Error location: [%s] \n" +
                                      "Error encountered is: [%s]",
                                      compatibilityResult.getErrorLocation(),
                                      compatibilityResult.getErrorMessage());
        LOG.error(errMsg);
        throw new IncompatibleSchemaException(errMsg);
    }
```

schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/SchemaMetadataStorable.java
```diff
@@ -80,6 +81,12 @@
      */
     private SchemaCompatibility compatibility = SchemaCompatibility.DEFAULT_COMPATIBILITY;
 
+
+    /**
+     * Validation level of this schema instance
+     */
+    private SchemaValidationLevel validation = SchemaValidationLevel.DEFAULT_VALIDATION_LEVEL;
```
satishd: Can it be renamed to `validationLevel` and respective get/set and schema field name etc?

schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/DefaultSchemaRegistry.java
michaelandrepearce: yeah was thinking the same when i did the code.

schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/SchemaMetadataStorable.java
michaelandrepearce: agreed, this is in a few places ill go through an update all

michaelandrepearce:
@satishd thanks for the review, is what i needed, agreed i need to add tests, will address the other bits.

michaelandrepearce:
updated pr, address all concerns, still have test case to add.

satishd:
@michaelandrepearce Overall LGTM, +1 once tests are added.

michaelandrepearce:
@satishd exposed method in client so service and client interfaces don't drift.
Add TEST case for the expose api method.
Add TEST case that checks both ValidationLevel = ALL, and ValidationLevel = LATEST

schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/DefaultSchemaRegistry.java
```diff
@@ -368,6 +351,25 @@ public SchemaMetadataInfo getSchemaMetadata(String schemaName) {
...
+    public SchemaMetadataInfo updateSchemaMetadata(String schemaName, SchemaMetadata schemaMetadata) {
+        if (!schemaName.equals(schemaMetadata.getName())) {
+            throw new IllegalArgumentException("schemaName must match the name in schemaMetadata");
+        }
+        SchemaMetadataStorable givenSchemaMetadataStorable = new SchemaMetadataStorable();
+        givenSchemaMetadataStorable.setName(schemaName);
+
+        SchemaMetadataStorable schemaMetadataStorable = storageManager.get(givenSchemaMetadataStorable.getStorableKey());
+        if (schemaMetadataStorable != null) {
+            schemaMetadataStorable = SchemaMetadataStorable.updateSchemaMetadata(schemaMetadataStorable, schemaMetadata);
```
satishd: @michaelandrepearce Is user allowed to upgrade compatibility level from `LATEST` to `ALL` which may result in incompatible versions of the existing schemas?

michaelandrepearce: Yes. This is the case also in confluence. Thus the need to be able to delete/disable schemas which imo is independent of life cycle management.

michaelandrepearce: *confluent

satishd: That does not look to be right design choice in other SR unless I am missing something here. System should not allow user to overwrite the schema versions into incompatible state from earlier.

michaelandrepearce: @satishd this is reality of what end users such as my org need.

michaelandrepearce: Along with the disable feature

michaelandrepearce: E.g. We can add that check only after ability to support disable, but there needs an ability to switch. Which withou disable as noted you would break compatibility if going latest to all.

satishd: @michaelandrepearce If this is really needed then it is better to set validation level at a specific version and that should be valid from the version it has modified to subsequent versions.

michaelandrepearce: @satishd  You're missing the point of the need and the benefits of disable(delete). This is one of the reasons, and why it was so called for in confluents and added there.

michaelandrepearce: This would be a non issue if you didn't try join the two requirements of deletion(disable) with the want to provide also lifecycle management

michaelandrepearce: Also to note, Without supporting this we won't be able to import schemas from confluent.

satishd: @michaelandrepearce We understand that disable/archive/delete should be added but it must be part of life cycle management of schema/versions. 
My only worry is it should not break existing versions into incompatible set of versions and then trying to stitch with disable/delete specific versions.  We can discuss more on this in #177 design.

michaelandrepearce: So if you can disable a schema essentially you're saying: no longer include this specific schema in validation.

As such you can then put this precheck in, as someone will have the ability to disable schemas if they really need to. It maybe just one schema in 20 versions just needs disabling, as such saying validate all from x would not be right.

Without the ability to disable validation of schemas we cannot add this pre check. We can add it later once disable is resolved.

Can we merge this therefor with the idea we retro add the back check once we have the ability to disable.

Re 177 I see no movement on alternate options.

satishd: @michaelandrepearce I have already mentioned [here](https://github.com/hortonworks/registry/pull/175#pullrequestreview-48816224) that this can be merged in after resolving conflicts. I would like to have lifecycle management of schema/versions #177 address the scenarios you have mentioned. Initial design will be shared in next 3-4 days, community can discuss to enhance that to address different usecases.

michaelandrepearce:
@satishd just rebased.

satishd:
@michaelandrepearce Thanks for addressing #171 with this PR.

Closes #171

PR #180, opened 2017-07-10
Kafka Avro Serdes

michaelandrepearce:
Added Kafka Avro Serde which is needed for Kafka Streams API, etc.

This is a PR for squashed version of PR 64.


michaelandrepearce:
@satishd re-created the squashed pr branch.

PR #185, opened 2017-07-14
ISSUE-184 Missing column in schema_metadata_info tables.

michaelandrepearce:
Add to create scripts update to add column if not existing. 
n.b. postgres statement is 9.6 compatible only.


satishd:
This fix does not suffice, I am already working on this. I am testing this out now and I will raise PR soon.

PR #186, opened 2017-07-14
ISSUE-184 Resolved missing columns in database and generating respecti…

### LINKED ISSUE #184: Missing column in schema_metadata_info tables.
Below error is thrown when a schema is inserted

```
ERROR [2017-07-14 05:13:40,609] com.hortonworks.registries.schemaregistry.webservice.SchemaRegistryResource: Error encountered while adding
104 ! org.mariadb.jdbc.internal.util.dao.QueryException: Unknown column 'validationLevel' in 'field list'
105 ! Query is: INSERT INTO schema_metadata_info (`id`, `name`, `description`, `schemaGroup`, `type`, `compatibility`, `validationLevel`, `time
106 ! at org.mariadb.jdbc.internal.protocol.AbstractQueryProtocol.getResult(AbstractQueryProtocol.java:1114)
107 ! at org.mariadb.jdbc.internal.protocol.AbstractQueryProtocol.executeQuery(AbstractQueryProtocol.java:172)
108 ! at org.mariadb.jdbc.MariaDbClientPreparedStatement.executeInternal(MariaDbClientPreparedStatement.java:230)
109 ! ... 73 common frames omitted
110 ! Causing: java.sql.SQLSyntaxErrorException: (conn:99) Unknown column 'validationLevel' in 'field list'
```

satishd:
…ve values for validationLevel

Closes #184

Did not migration statements in sql scripts as it is different for different versions, that can be addressed later.

### DIFF
--- a/bootstrap/sql/mysql/create_tables.sql
+++ b/bootstrap/sql/mysql/create_tables.sql
@@ -14,15 +14,18 @@
 
 -- THE NAMES OF THE TABLE COLUMNS MUST MATCH THE NAMES OF THE CORRESPONDING CLASS MODEL FIELDS
 
+-- USE schema_registry;
+
 CREATE TABLE IF NOT EXISTS schema_metadata_info (
-  id            BIGINT AUTO_INCREMENT NOT NULL,
-  type          VARCHAR(256)          NOT NULL,
-  schemaGroup   VARCHAR(256)          NOT NULL,
-  name          VARCHAR(256)          NOT NULL,
-  compatibility VARCHAR(256)          NOT NULL,
-  description   TEXT,
-  evolve        BOOLEAN               NOT NULL,
-  timestamp     BIGINT                NOT NULL,
+  id              BIGINT AUTO_INCREMENT NOT NULL,
+  type            VARCHAR(256)          NOT NULL,
+  schemaGroup     VARCHAR(256)          NOT NULL,
+  name            VARCHAR(256)          NOT NULL,
+  compatibility   VARCHAR(256)          NOT NULL,
+  validationLevel VARCHAR(256)          NOT NULL, -- added in 0.3.1, table should be altered to add this column from earlier versions.
+  description     TEXT,
+  evolve          BOOLEAN               NOT NULL,
+  timestamp       BIGINT                NOT NULL,
   PRIMARY KEY (name),
   UNIQUE KEY (id)
 );
@@ -54,13 +57,13 @@ CREATE TABLE IF NOT EXISTS schema_field_info (
 );
 
 CREATE TABLE IF NOT EXISTS schema_serdes_info (
-  id                     BIGINT AUTO_INCREMENT NOT NULL,
-  description            TEXT,
-  name                   TEXT                  NOT NULL,
-  fileId                 TEXT                  NOT NULL,
-  serializerClassName    TEXT                  NOT NULL,
-  deserializerClassName  TEXT                  NOT NULL,
-  timestamp              BIGINT                NOT NULL,
+  id                    BIGINT AUTO_INCREMENT NOT NULL,
+  description           TEXT,
+  name                  TEXT                  NOT NULL,
+  fileId                TEXT                  NOT NULL,
+  serializerClassName   TEXT                  NOT NULL,
+  deserializerClassName TEXT                  NOT NULL,
+  timestamp             BIGINT                NOT NULL,
   PRIMARY KEY (id)
 );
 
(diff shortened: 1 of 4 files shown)


schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/SchemaMetadataStorable.java
```diff
@@ -224,8 +230,17 @@ public static SchemaMetadataStorable updateSchemaMetadata(SchemaMetadataStorable
         schemaMetadataStorable.setSchemaGroup(schemaMetadata.getSchemaGroup());
         schemaMetadataStorable.setName(schemaMetadata.getName());
         schemaMetadataStorable.setDescription(schemaMetadata.getDescription());
-        schemaMetadataStorable.setCompatibility(schemaMetadata.getCompatibility());
-        schemaMetadataStorable.setValidationLevel(schemaMetadata.getValidationLevel());
+
+        schemaMetadataStorable.setCompatibility(
```
michaelandrepearce: the if null check / defaulting is done already in schema meta data. 

https://github.com/hortonworks/registry/blob/master/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/SchemaMetadata.java

bootstrap/sql/mysql/create_tables.sql
```diff
@@ -14,15 +14,18 @@
...
-  compatibility VARCHAR(256)          NOT NULL,
-  description   TEXT,
-  evolve        BOOLEAN               NOT NULL,
-  timestamp     BIGINT                NOT NULL,
+  id              BIGINT AUTO_INCREMENT NOT NULL,
+  type            VARCHAR(256)          NOT NULL,
+  schemaGroup     VARCHAR(256)          NOT NULL,
+  name            VARCHAR(256)          NOT NULL,
+  compatibility   VARCHAR(256)          NOT NULL,
+  validationLevel VARCHAR(256)          NOT NULL, -- added in 0.3.1, table should be altered to add this column from earlier versions.
```
michaelandrepearce: why not make this a table alter statement as per my PR, this way it would handle upgrades, instead of making people manually have to work out what to do.

bootstrap/sql/postgresql/create_tables.sql
```diff
@@ -14,50 +14,53 @@
...
-  "description"   TEXT,
-  "evolve"        BOOLEAN               NOT NULL,
-  "timestamp"     BIGINT                NOT NULL,
-  UNIQUE("id","name")
+  "id"              SERIAL PRIMARY KEY,
+  "type"            VARCHAR(256) NOT NULL,
+  "schemaGroup"     VARCHAR(256) NOT NULL,
+  "name"            VARCHAR(256) NOT NULL,
+  "compatibility"   VARCHAR(256) NOT NULL,
+  "validationLevel" VARCHAR(256) NOT NULL, -- added in 0.3.1, table should be altered to add this column from earlier versions.
```
michaelandrepearce: ditto to comment in mysql

bootstrap/sql/mysql/create_tables.sql
michaelandrepearce: e.g. 

+IF NOT EXISTS( SELECT NULL
 +            FROM INFORMATION_SCHEMA.COLUMNS
 +           WHERE table_name = 'schema_metadata_info'
 +             AND column_name = 'validationLevel')  THEN
 +  ALTER TABLE `schema_metadata_info` ADD `validationLevel` VARCHAR(256) NOT NULL DEFAULT 'ALL';
 +END IF;

schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/SchemaMetadataStorable.java
satishd: When generated from jsons (which have null values)  with default constructor can  lead to nulls.

michaelandrepearce: If this is the case would it not be better to keep still the logic in schema metadata and move to the getter methods to have the null check instead of having to null check many times in code. (Or needing to remember to do so)

satishd: @michaelandrepearce  makes sense, will do.

bootstrap/sql/mysql/create_tables.sql
satishd: @michaelandrepearce I did not have those changes as you already have them in other PR. You can merge your changes in PR and mention in the comments what versions work fine for this script in the comments.

michaelandrepearce:
+1 thanks @satishd

Issue #189, opened 2017-07-18
Sanitize mysql create_tables scripts


arunmahadevan:
In many columns we have defined column types as `varchar(256)`. This can potentially throw error during table creation when using utf-8 charset. Revisit the column definitions and use sane limits with max = 255 instead of 256.

Exception in thread "main" java.sql.SQLSyntaxErrorException: (conn:12499) Specified key was too long; max key length is 767 bytes

satishd:
Right, it should be done for below reasons
- utf-8 requires 3 bytes for each characters, 3*256 = 768 and column key limit on innodb is 767
- utf8mb4 requires 4 bytes, so max length can be 191 instead of 255
- 256 requires 2 bytes which is unnecessary.

I am not sure why these were created with 256 though. We can have this as 255 and when user uses utf8mb4 then they can change their scripts to have respective length restriction as 191.

satishd:
#190

PR #190, opened 2017-07-18
ISSUE-189 Sanitize mysql create_tables scripts

satishd:


arunmahadevan:
+1, are there any limits with postgres as well ? as long as its 255 or less we are good.

michaelandrepearce:
should this be possibly pushed into the 0.3.0 RC ?

satishd:
@arunmahadevan 
> +1, are there any limits with postgres as well ? as long as its 255 or less we are good.

- maximum size for a field is 1GB which is again configurable
- One limitation is that indexes can not be created on columns longer than about 2,000 characters.

More details can be found [here](http://wiki.postgresql.org/wiki/FAQ#What_is_the_maximum_size_for_a_row.2C_a_table.2C_and_a_database.3F) , so having 255 is fine and it is consistent.

satishd:
@michaelandrepearce good to be part of 0.3.0 RC.

satishd:
Closes #189

PR #212, opened 2017-07-30
ISSUE-211 Adding missing APIs for Confluent SR and a fix in returning right id for version.

satishd:
- Added more APIs for Confluent SR.
- Fixed returning version id instead of metadata id.
- Added tests for APIs including a stress test.

Tested and verified with MySql storage.


satishd:
Closes #211

michaelandrepearce:
@satishd  i note theres a 0.3.1.1 branch, where there was a merge from master 17days ago, as such its got the bug, and prob this should be merged into that branch if a tag is being made soon there. else it will be broken.

Issue #213, opened 2017-07-31
POJOs should ignore fields which do not have respective fields for backward compatibility by clients.


satishd:
Currently, SchemaMetadata and other pojos do not ignore unrecognizable fields. This should be done to provide backward compatibility by clients when there are no fields recognized on server side.

satishd:
#215

PR #215, opened 2017-08-01
ISSUE-213 POJOs are made to be backward compatible

satishd:
This allows schema registry clients with earlier versions to send old POJOs to new versions of SchemaRegistry server and deserialization will not look for missing fields in respective POJOs SchemaRegsitry server.


satishd:
Closes #213

PR #217, opened 2017-08-02
Fix registerSubjectVersion method name to lookupSubjectVersion in Con…

satishd:
…fluentSchemaRegistryCompatibleResource.


michaelandrepearce:
+1

Issue #224, opened 2017-08-09
Return respective errors for Confluent's /subjects/{subject}/versions/{versionId} API.


satishd:
This is to address API contract for errors mentioned in [API doc](http://docs.confluent.io/current/schema-registry/docs/api.html).

PR #225, opened 2017-08-09
ISSUE-224 Return respective error messages for different subject and version not found cases.

satishd:
- Return respective error messages for different subject and version not found cases. 
- Added respective tests.

### DIFF
--- a/schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/ConfluentSchemaRegistryCompatibleResource.java
+++ b/schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/ConfluentSchemaRegistryCompatibleResource.java
@@ -164,7 +164,7 @@ public class ConfluentSchemaRegistryCompatibleResource extends BaseRegistryResou
 
     @GET
     @Path("/subjects/{subject}/versions")
-    @ApiOperation(value = "Get all registered subjects",
+    @ApiOperation(value = "Get all schema versions of given subject",
             response = Integer.class, responseContainer = "Collection", tags = OPERATION_GROUP_CONFLUENT_SR)
     @Timed
     public Response getAllVersions(@ApiParam(value = "subject", required = true)
@@ -190,7 +190,7 @@ public class ConfluentSchemaRegistryCompatibleResource extends BaseRegistryResou
 
     @GET
     @Path("/subjects/{subject}/versions/{versionId}")
-    @ApiOperation(value = "Get all registered subjects",
+    @ApiOperation(value = "Get the schema information for given subject and versionId",
             response = Integer.class, responseContainer = "Collection", tags = OPERATION_GROUP_CONFLUENT_SR)
     @Timed
     public Response getSchemaVersion(@ApiParam(value = "subject", required = true)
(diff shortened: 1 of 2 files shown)


satishd:
Closes #224

michaelandrepearce:
@satishd I will look tomorrow when in work. I know I didn't handle / map all the errors as the confluent serdes makes no distinction really, but it is a good thing to sort

schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/ConfluentSchemaRegistryCompatibleResource.java
```diff
@@ -201,18 +201,42 @@ public Response getSchemaVersion(@ApiParam(value = "subject", required = true)
                                              String versionId) {
         Response response;
         try {
+            SchemaVersionInfo schemaVersionInfo = null;
 
-            SchemaVersionInfo schemaVersionInfo = "latest".equals(versionId)
-                    ? schemaRegistry.getLatestSchemaVersionInfo(subject)
-                    : schemaRegistry.getSchemaVersionInfo(Long.valueOf(versionId));
+            if ("latest".equals(versionId)) {
```
michaelandrepearce: we should probably add some versionId validation and throw invalidVersionException() if not valid input range.
see here for logic of the id validation https://github.com/confluentinc/schema-registry/blob/8eb664dbc84b1c2db3666fa0771eeb0e0909f892/core/src/main/java/io/confluent/kafka/schemaregistry/rest/VersionId.java

michaelandrepearce: this should be on all paths that take versionId

schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/ConfluentSchemaRegistryCompatibleResource.java
```diff
@@ -201,18 +201,42 @@ public Response getSchemaVersion(@ApiParam(value = "subject", required = true)
...
+                response = WSUtils.respondEntity(schemaVersionEntry, Response.Status.OK);
+            }
         } catch (SchemaNotFoundException ex) {
-            LOG.error("No schema version found with id [{}]", versionId, ex);
-            response = versionNotFoundError();
+            LOG.error("No schema found with subject [{}]", subject, ex);
+            response = subjectNotFoundError();
         } catch (Exception ex) {
             LOG.error("Encountered error while retrieving all subjects", ex);
             response = serverError();
```
michaelandrepearce: should probably return RestSchemaRegistryException for these, wdyt?

michaelandrepearce:
I don't see any issues with what is here, e.g. nothing incorrect. 

There is probably a few more cases we should add if we want exception handling to match also identical to confluent. I have added some comments where should be thought about. This though can be in another PR don't see why all has to be done in one. So I'm a +1 for this to merge

schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/ConfluentSchemaRegistryCompatibleResource.java
satishd: No plans to address these as they are specific errors form storage etc. Users would anyway see the error message on the reason of failure of the invocation in their error message. Idea of this PR is to handle respective errors for no schema/version found.

satishd:
Thanks @michaelandrepearce for reviewing, addressed/replied to your comments.

Issue #220, opened 2017-08-04
Schema Version Diff UI


shahsank3t:
Provide a UI to check the diff between any two versions of a particular schema.

satishd:
+1

shahsank3t:
https://github.com/hortonworks/registry/pull/229 fixes this issue

PR #229, opened 2017-08-11
Issue-220-UI Schema Version Diff UI

joylyn:
UI changes for #220

### DIFF
--- a/webservice/src/main/resources/app/scripts/containers/Registry-Services/SchemaRegistry/SchemaVersionDiff.jsx
+++ b/webservice/src/main/resources/app/scripts/containers/Registry-Services/SchemaRegistry/SchemaVersionDiff.jsx
@@ -1,0 +1,170 @@
+/**
+  * Copyright 2017 Hortonworks.
+  *
+  * Licensed under the Apache License, Version 2.0 (the "License");
+  * you may not use this file except in compliance with the License.
+  * You may obtain a copy of the License at
+  *   http://www.apache.org/licenses/LICENSE-2.0
+  * Unless required by applicable law or agreed to in writing, software
+  * distributed under the License is distributed on an "AS IS" BASIS,
+  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+**/
+
+import React, {Component} from 'react';
+import ReactDOM from 'react-dom';
+import _ from 'lodash';
+import Select from 'react-select';
+import {Button, ButtonGroup} from 'react-bootstrap';
+import Utils from '../../../utils/Utils';
+import {diffview} from '../../../libs/jsdifflib/diffview';
+import {difflib} from '../../../libs/jsdifflib/difflib';
+
+export default class SchemaVersionDiff extends Component {
+  constructor(props) {
+    super(props);
+    let versionsArr = Utils.sortArray(this.props.schemaObj.versionsArr.slice(), 'versionId', false);
+    let newTextOptions = versionsArr;
+    let baseTextOptions = versionsArr;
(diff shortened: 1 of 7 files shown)


webservice/src/main/resources/app/scripts/containers/Registry-Services/SchemaRegistry/SchemaVersionDiff.jsx
```diff
@@ -0,0 +1,159 @@
...
+      opcodes: opcodes,
+      baseTextName: baseTextVersion ? ("V " + baseTextVersion.versionId) : '',
+      newTextName: newTextVersion ?  ("V " + newTextVersion.versionId) : '',
+      viewType: 1
+    }));
+
+  }
+
+  render() {
+    const jsonoptions = {
```
shahsank3t: I believe we are not referring codemirror or this jsonoptions anywhere so it would be good to get rid of those.

webservice/src/main/resources/app/scripts/containers/Registry-Services/SchemaRegistry/SchemaVersionDiff.jsx
```diff
@@ -0,0 +1,159 @@
...
+import Select from 'react-select';
+import Utils from '../../../utils/Utils';
+import ReactCodemirror from 'react-codemirror';
+import CodeMirror from 'codemirror';
+import 'codemirror/mode/javascript/javascript';
+import jsonlint from 'jsonlint';
+import lint from 'codemirror/addon/lint/lint';
+import {diffview} from '../../../libs/jsdifflib/diffview';
+import {difflib} from '../../../libs/jsdifflib/difflib';
+import '../../../libs/jsdifflib/diffview.css';
```
shahsank3t: We are importing all the css files into main.js file, can you please add this over there.

shahsank3t:
+1 looks good to me.

shahsank3t:
@satishd @harshach @michaelandrepearce 
Attaching an image below to show the compare UI screen after applying this patch.

screen shot 2017-08-11 at 12 27 46 pm

You guys can play around using the PR and once approved, I'll merge it in.

satishd:
@shahsank3t This is good to start with, but it good to have an option to see the diff side by side also.

shahsank3t:
@satishd sure, we'll try out side-by-side diff(option to choose this or single diff) and see how it goes.

harshach:
+1 on side by side view of diff

joylyn:
Committed changes for Side by Side view of Diff.

shahsank3t:
This looks good @joylyn 
A couple of suggestions:
1. Can we decrease the padding between the label and version drop down?
2. Move the buttons on the same line of version drop down (right side).
3. Rename "side by side diff" to "Split" and "Inline diff" to "Unified" and by default select "Unified" (taking reference from github)

joylyn:
@shahsank3t Thanks for the suggestions. Addressed the same.

joylyn:
@shahsank3t @satishd Made changes to show the version description and reduced size of drop-downs.

shahsank3t:
@satishd Sure, I'll try out with more versions and look into the issues you mentioned.

joylyn:
@satishd  To avoid multiple clicks, I can show all the version numbers in the dropdowns. Hence if we select a higher version as the base version and an older version on the right we get the reverse diff. Is this the correct result for this case ?

joylyn:
I have committed some changes to work as follows
- When we select a higher version as base, the new version will change to the next higher version. Similarly if we select same version, it should change . Hence in this case, we exclude the current version from the base drop-down and initial version from the right drop-down respectively.

joylyn:
Committed changes to show all versions in the dropdown.

PR #233, opened 2017-08-18
ISSUE-177 Schema version lifecycle state management implementation.

satishd:
- Implementation of schema version life cycle state management, doc is located [here](https://docs.google.com/document/d/1qi3o9ADxkySZn-PZJiWTyfcvdBraW-dHGwMRm-xZeyA/edit?usp=sharing)

- Added respective unit and integration tests
- Minor javadoc fixes

### DIFF
--- a/bootstrap/sql/mysql/create_tables.sql
+++ b/bootstrap/sql/mysql/create_tables.sql
@@ -38,11 +38,14 @@ CREATE TABLE IF NOT EXISTS schema_version_info (
   version          INT                   NOT NULL,
   schemaMetadataId BIGINT                NOT NULL,
   timestamp        BIGINT                NOT NULL,
+  state            TINYINT               NOT NULL,
   name             VARCHAR(255)          NOT NULL,
   UNIQUE KEY (id),
   UNIQUE KEY `UK_METADATA_ID_VERSION_FK` (schemaMetadataId, version),
   PRIMARY KEY (name, version),
-  FOREIGN KEY (schemaMetadataId, name) REFERENCES schema_metadata_info (id, name) ON DELETE CASCADE ON UPDATE CASCADE
+  FOREIGN KEY (schemaMetadataId, name) REFERENCES schema_metadata_info (id, name)
+    ON DELETE CASCADE
+    ON UPDATE CASCADE
 );
 
 CREATE TABLE IF NOT EXISTS schema_field_info (
@@ -53,7 +56,20 @@ CREATE TABLE IF NOT EXISTS schema_field_info (
   fieldNamespace   VARCHAR(255),
   type             VARCHAR(255)          NOT NULL,
   PRIMARY KEY (id),
-  FOREIGN KEY (schemaInstanceId) REFERENCES schema_version_info (id) ON DELETE CASCADE ON UPDATE CASCADE
+  FOREIGN KEY (schemaInstanceId) REFERENCES schema_version_info (id)
+    ON DELETE CASCADE
+    ON UPDATE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS schema_version_state (
+  id              BIGINT AUTO_INCREMENT NOT NULL,
+  schemaVersionId BIGINT                NOT NULL,
+  stateId         TINYINT               NOT NULL,
+  sequence        INT                   NOT NULL,
+  timestamp       BIGINT                NOT NULL,
+  details         VARCHAR(255),
+  PRIMARY KEY (schemaversionId, stateId, sequence),
+  UNIQUE KEY (id)
 );
 
 CREATE TABLE IF NOT EXISTS schema_serdes_info (
--- a/schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
+++ b/schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
@@ -47,6 +47,7 @@ import com.hortonworks.registries.schemaregistry.serde.SnapshotSerializer;
 import com.hortonworks.registries.schemaregistry.serde.pull.PullDeserializer;
 import com.hortonworks.registries.schemaregistry.serde.pull.PullSerializer;
 import com.hortonworks.registries.schemaregistry.serde.push.PushDeserializer;
+import com.hortonworks.registries.schemaregistry.state.SchemaLifecycleException;
 import org.apache.commons.io.IOUtils;
 import org.glassfish.jersey.SslConfigurator;
 import org.glassfish.jersey.client.ClientConfig;
@@ -233,7 +234,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                 return doGetSchemaVersionInfo(key);
             }
         },
-                ((Number) configuration.getValue(Configuration.SCHEMA_VERSION_CACHE_SIZE.name())).longValue(),
+                ((Number) configuration.getValue(Configuration.SCHEMA_VERSION_CACHE_SIZE.name())).intValue(),
                 ((Number) configuration.getValue(Configuration.SCHEMA_VERSION_CACHE_EXPIRY_INTERVAL_SECS.name())).longValue());
 
         SchemaMetadataCache.SchemaMetadataFetcher schemaMetadataFetcher = createSchemaMetadataFetcher();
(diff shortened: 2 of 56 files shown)


schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
```diff
@@ -614,6 +620,81 @@ private static String encode(String schemaName) {
...
+    }
+
+    @Override
+    public void startSchemaVersionReview(Long schemaVersionId) throws SchemaNotFoundException, SchemaLifecycleException {
+        fetchSchemaVersionState(schemaVersionId, "startReview");
+
+    }
+
+    private boolean fetchSchemaVersionState(Long schemaVersionId,
+                                            String operation) throws SchemaNotFoundException, SchemaLifecycleException {
```
michaelandrepearce: operation should probably be an enum.

michaelandrepearce: on that note, it seems we switch between tinyint and string, should keep to one or the other?

bootstrap/sql/mysql/create_tables.sql
```diff
@@ -38,11 +38,14 @@ CREATE TABLE IF NOT EXISTS schema_version_info (
   version          INT                   NOT NULL,
   schemaMetadataId BIGINT                NOT NULL,
   timestamp        BIGINT                NOT NULL,
+  state            TINYINT               NOT NULL,
```
michaelandrepearce: Really need upgrade scripts, the amount of changes in the db now, is getting dangerous for any users with existing data, having to custom upgrade their DB without any upgrade scripts that also are tested.

michaelandrepearce:
I cannot find logic or tests that checks/ignores disabled and archived schema's from schema validation.

schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/SchemaRegistryResource.java
```diff
@@ -606,6 +608,179 @@ public Response getSchemaVersionById(@ApiParam(value = "version identifier of th
...
+        } catch (Exception ex) {
+            LOG.error("Encountered error while getting schema version lifecycle states", ex);
+            response = WSUtils.respond(Response.Status.INTERNAL_SERVER_ERROR, CatalogResponse.ResponseMessage.EXCEPTION, ex.getMessage());
+        }
+
+        return response;
+    }
+
+    @POST
+    @Path("/schemas/versionsById/{id}/states/enable")
```
michaelandrepearce: why is states pluralised, surely a schema version can have only a singular state.

e.g. should this not be:
/schemas/versionsById/{id}/state/enable

schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/SchemaRegistryResource.java
```diff
@@ -606,6 +608,179 @@ public Response getSchemaVersionById(@ApiParam(value = "version identifier of th
...
+        } catch (Exception ex) {
+            LOG.error("Encountered error while getting schema version with id [{}]", versionId, ex);
+            response = WSUtils.respond(Response.Status.INTERNAL_SERVER_ERROR, CatalogResponse.ResponseMessage.EXCEPTION, ex.getMessage());
+        }
+
+        return response;
+    }
+
+    @POST
+    @Path("/schemas/versionsById/{id}/states/disable")
```
michaelandrepearce: ditto above comment, won't repeat any further.

schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/SchemaRegistryResource.java
```diff
@@ -606,6 +608,179 @@ public Response getSchemaVersionById(@ApiParam(value = "version identifier of th
...
+            response = WSUtils.respond(Response.Status.INTERNAL_SERVER_ERROR, CatalogResponse.ResponseMessage.EXCEPTION, ex.getMessage());
+        }
+
+        return response;
+    }
+
+
+    @POST
+    @Path("/schemas/versionsById/{id}/states/custom")
+    @ApiOperation(value = "Runs the custom state execution for schema version identified by the given version id",
```
michaelandrepearce: I assume there are many custom state options, how is one identified?

bootstrap/sql/mysql/create_tables.sql
michaelandrepearce: Id be really pleased to see https://github.com/hortonworks/registry/pull/232 getting put in before these changes, and then let it manage this upgrade

michaelandrepearce:
@satishd looks great as a first cut, have left some comments inline and also in general. 
Two key bits i have is
* ignoring validation for disabled/archived
* how custom states are configured/stored and avoid conflict with pre-defined.

I assume UI stuff will be complete separate PR and this is to get the core scaffolding down.

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/SchemaVersion.java
```diff
@@ -37,6 +39,14 @@ public SchemaVersion(String schemaText, String description) {
         this.schemaText = schemaText;
     }
 
+    public SchemaVersion(String schemaText, String description, Byte initialState) {
+        this.description = description;
+        this.schemaText = schemaText;
+        if(initialState != null) {
```
michaelandrepearce: Style of defaulting differs to other places such as https://github.com/satishd/registry-1/blob/669b66f1ce0597398c5dc277c7099b95dd9ac957/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/SchemaVersionInfo.java

should follow standard pattern in default setting a value across the project, all ways are valid, but really helps with consistency of code

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/SchemaVersion.java
```diff
@@ -28,6 +29,7 @@
 
     private String description;
     private String schemaText;
+    private Byte initialState = SchemaVersionLifecycleStates.ENABLED.id();
```
michaelandrepearce: should this not be just, stateId

bootstrap/sql/mysql/create_tables.sql
satishd: That is the plan, #232 will be merged in before this.

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/SchemaVersion.java
satishd: I made it explicit that it is the initial state of the version being added. I am fine with either `stateId` or `initialStateId`.

schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/SchemaRegistryResource.java
satishd: Sure, that makes sense.

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
satishd: IMHO, we do not really need a table as the statemachine is always stored as part of the code. 
States are not customizable. I plan to expose an API for statemachine so that UI can use that to build respective actions. 
You raised a valid concern of having strings as operations in REST APIs. This was done to give better expressive REST APIs. These strings were planned to be taken from the state machine api response. But I agree that it is better to be exposed with ids(which are bytes) instead of strings.

satishd:
@michaelandrepearce Thanks for reviewing this PR.

Unit tests for different states are added in [SchemaVersionLifecycleStatesTest](https://github.com/hortonworks/registry/pull/233/files#diff-4fa115ec535a592ff1cd224a00df3d05). It seems that test is disabled now, I will enable it in next commit.

> ignoring validation for disabled/archived

Validation is done only for enabled schemas. Let me know if I am missing anything here.

>how custom states are configured/stored and avoid conflict with pre-defined.

 [SchemaVersionLifecycleStates.Registry#register](https://github.com/hortonworks/registry/pull/233/files#diff-8d9b5c996e68c93d1b31c4ba40044fd0) does not allow states with duplicate state ids. [SchemaVersionLifecycleState#id()](https://github.com/hortonworks/registry/pull/233/files#diff-0ea736c3e0e93412c797bf7d02ae52be) explains reserved id range. I will add a sample/test on how to build custom state and how that can be used. 

> I assume UI stuff will be complete separate PR and this is to get the core scaffolding down.

Right, UI will be a separate PR. Need to add an API for getting statemachine from the backend which will be used by UI.

> Also as noted i think getting the flyway db management PR in first, would be a great plus to this work removing end user risk going forwards.

Right, Plan is to merge #232 before this PR.

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
michaelandrepearce: Is this method interface changing then?

michaelandrepearce:
Thanks @satishd 

I think this should be in the documents not just in code:
> SchemaVersionLifecycleState#id() explains reserved id range. I will add a sample/test on how to build custom state and how that can be used.

Spotted this now (all code is moved) and agreed logic looks right
> Validation is done only for enabled schemas. Let me know if I am missing anything here.

+1

satishd:
Thanks @michaelandrepearce for your review. I will add sample about adding custom state along with the required changes for UI related APIs. I may have minor tweaks in current API which can be discussed then.

Issue #238, opened 2017-08-30
AvroSchemaProvider does not take into account different String representations when canonicalizing


radai-rosenblatt:
avro has a magic property "avro.java.string" that determines the type of a string property in generated java code (java.lang.String vs Utf8 etc) - see https://issues.apache.org/jira/browse/AVRO-803 and https://stackoverflow.com/questions/25118727/how-to-generate-fields-of-type-string-instead-of-charsequence-using-avro

changing the value of this property on a schema, while having no effect on compatibility as far as wire format is concerned (both json and binary) can definitely break code compiled from/against this schema (at least in java).

i think then it should be at least included in the fingerprint, if not also be made part of the validation?

satishd:
@radai-rosenblatt Fingerprint is used in finding out whether the same effective schema w.r.t compatibility already exists in registry for the same schema name before it is added as new version. When user adds new schema with different avro.java.string types, it should not be added as a new version as it does not change much w.r.t compatibility. But validation can be added to check whether avro.java.string is set with right values.

PR #240, opened 2017-09-06
ISSUE-238 : Make AvroSchemaProvider honour string representations when canonicalising

raju-saravanan:
This pull request addresses the issue: #238

### DIFF
--- a/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/avro/AvroSchemaProvider.java
+++ b/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/avro/AvroSchemaProvider.java
@@ -23,6 +23,8 @@ import com.hortonworks.registries.schemaregistry.SchemaFieldInfo;
 import com.hortonworks.registries.schemaregistry.errors.InvalidSchemaException;
 import com.hortonworks.registries.schemaregistry.errors.SchemaNotFoundException;
 import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
+import org.apache.commons.lang3.EnumUtils;
 import org.codehaus.jackson.JsonNode;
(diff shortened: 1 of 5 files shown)


schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/avro/AvroSchemaProvider.java
```diff
@@ -106,14 +108,24 @@ public String normalize(Schema schema) throws IOException {
...
         String fullName = schema.getFullName();
         switch (schemaType) {
-            default: // boolean, bytes, double, float, int, long, null, string
+            default: // boolean, bytes, double, float, int, long, null
                 return appendable.append('"').append(schemaType.getName()).append('"');
 
+            case STRING:
+                String stringType = schema.getProp(GenericData.STRING_PROP);
+                if ( stringType == null )
+                    stringType = schemaType.getName();
```
satishd: When `GenericData.STRING_PROP` does not exist, the value is considered as `Utf8`. 
string type
- without `GenericData.STRING_PROP`
- with `GenericData.STRING_PROP` as Utf8
should generate same finger print.

But the current logic generates different strings in the above scenarios. So, it is better to handle default value for `GenericData.STRING_PROP` and generate equivalent strings. One way to fix this is always have `schemaType.getName()` and append effective value for `GenericData.STRING_PROP` property.
Generated string can be of the form like `{"string", "$javaType"}`, javaType can be CharSequence, String and Utf8.

Good to add tests to check whether it generates equivalent fingerprints for equivalent effective types respectively.

satishd:
@raju-saravanan After thinking through the reported issue more , I realized that these changes may not be really needed except for having validation of avro.java.string with right values. Put a comment in issue #238 and want to hear others opinions.

PR #246, opened 2017-09-12
ISSUE-242 State change implementation

satishwaghela:
UI Implementation for schema version state change.

### DIFF
--- a/webservice/src/main/resources/app/scripts/containers/Registry-Services/SchemaRegistry/SchemaRegistryContainer.jsx
+++ b/webservice/src/main/resources/app/scripts/containers/Registry-Services/SchemaRegistry/SchemaRegistryContainer.jsx
@@ -28,9 +28,11 @@ import {
     PanelGroup,
     Panel,
     Modal,
-    Pagination
+    Pagination,
+    OverlayTrigger,
+    Popover
 } from 'react-bootstrap';
-import Utils from '../../../utils/Utils';
+import Utils, {StateMachine} from '../../../utils/Utils';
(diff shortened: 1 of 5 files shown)


shahsank3t:
@satishd Can you please try this PR and provide feedback if any for the UI. Thanks in advance.

satishd:
This UI looks functional for basic state transformations. 
We have complete state trail for each version in the backend which we may want to expose to user. We may also want to show state flow with graphical state flow diagram so that users can easily understand the current state flow of schema registry. It may be useful for users when SR is started with customized state flow. I will raise Github issues for these after finalizing on how we want to expose state transition in UI.

webservice/src/main/resources/app/scripts/containers/Registry-Services/SchemaRegistry/SchemaRegistryContainer.jsx
```diff
@@ -63,6 +65,60 @@ CodeMirror.registerHelper("lint", "json", function(text) {
...
+    this.setState({edit: true}, () => {});
+  }
+  render(){
+    const {edit} = this.state;
+    const {StateMachine, version, showEditBtn} = this.props;
+    const transitions = StateMachine.getTransitionStateOptions(version.stateId);
+    const currentState = StateMachine.getStateById(version.stateId).name;
+    let comp;
+    if(edit){
+      comp = <div style={{"margin-top": "5px"}}>
```
shahsank3t: nit: this should be "marginTop"

shahsank3t:
@satishwaghela once the comment is addressed, +1 from me

PR #258, opened 2017-09-26
ISSUE-257: Missing favicon returns 404 error

joylyn:
UI change for #257


shahsank3t:
@satishd No, it's not expected to have any files from `public/` directory to be shown in git.
It seems like `public/assets/index.html` is already into our codebase which shouldn't be.

@joylyn could you please remove the `public` directory and verify that nothing from the public is shown into git. Thanks.

PR #270, opened 2017-10-10
ISSUE-269: Have a common storage module across streamline and registry

arunmahadevan:
Bringing in changes from streamline to maintain a common storage module across streamline and registry.


arunmahadevan:
@raju-saravanan can you add the oracle scripts on top of this for oracle support and test it out?

raju-saravanan:
@arunmahadevan sure, I will test it out for oracle.

storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/provider/sql/statement/DefaultStorageDataTypeContext.java
```diff
@@ -0,0 +1,167 @@
...
+            default:
+                throw new IllegalArgumentException("Not supported type: " + type);
+        }
+    }
+
+    protected Object getJavaObject(Class columnJavaType, String columnLabel, ResultSet resultSet) throws SQLException {
+        if (columnJavaType.equals(String.class)) {
+            return resultSet.getString(columnLabel);
+        } else if (columnJavaType.equals(Integer.class)) {
+            return resultSet.getInt(columnLabel);
```
raju-saravanan: We have a columnJavaType which maps to type "Byte", can you handle that case?

raju-saravanan:
@arunmahadevan I have tested your changes against Oracle (12c and 11g), registry is working as expected.

arunmahadevan:
@raju-saravanan addressed the comments and verified with mysql.

In postgres i am getting an error with type mismatch. But this seems to be an existing issue. I will track it in a separate issue.

raju-saravanan:
+1 , LGTM

common/src/main/java/com/hortonworks/registries/common/util/FileStorage.java
```diff
@@ -20,13 +20,13 @@
 import java.util.Map;
 
 /**
- * Interface abstracting the upload and download of files (like parser jars, split/join or custom processor jars or any other files) for IoTaS.
- * IoTaS will provide a default file system based implementation which can be swapped by another implementation using fileStorageConfiguration
- * property in the iotas.yaml
+ * Interface abstracting the upload and download of files (like parser jars, split/join or custom processor jars or any other files) for Streamline.
+ * Streamline will provide a default file system based implementation which can be swapped by another implementation using fileStorageConfiguration
```
michaelandrepearce: Can we remove reference to streamline

common/src/main/java/com/hortonworks/registries/common/util/FileStorage.java
```diff
@@ -20,13 +20,13 @@
...
+ * property in the streamline.yaml
  */
 public interface FileStorage {
 
-    String DEFAULT_DIR = "/tmp/iotas-files";
+    String DEFAULT_DIR = "/tmp/streamline-files";
```
michaelandrepearce: should be name registry or something not streamline

common/src/main/java/com/hortonworks/registries/common/util/FileStorage.java
satishd: Right. Can we remove any streamline related information like parser jar or split/join etc mentioned in this description? Lets not even say Registry as this is common module which can be used in any component.

common/src/main/java/com/hortonworks/registries/common/util/FileStorage.java
satishd: Do no use either registry or streamline but it can be like `/tmp/uploaded-files` or some other valid name.

storage/core/src/main/java/com/hortonworks/registries/storage/util/StorageUtils.java
```diff
@@ -64,11 +73,44 @@ public static void ensureUnique(Storable storable,
     @SuppressWarnings("unchecked")
     public static Collection<Class<? extends Storable>> getStreamlineEntities() {
```
satishd: @arunmahadevan Can you rename this to `getStorableEntities` or any other valid name? Lets not use streamline or registry specific names as it is common module.

arunmahadevan:
@michaelandrepearce  @satishd, addressed review comments.

satishd:
@michaelandrepearce I will merge this change once you have final review on the addressed comments.

PR #286, opened 2017-10-24
ISSUE-273: Transaction support in storage layer.

raju-saravanan:
- Transaction support in storage layer
- Made registry storage transactional using these new APIs.


storage/core/src/main/java/com/hortonworks/registries/storage/TransactionalStorageManager.java
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+package com.hortonworks.registries.storage;
+
+import com.hortonworks.registries.storage.exception.StorageException;
+
+public interface TransactionalStorageManager extends StorageManager {
```
satishd: We can have this interface as `TransactionManager` or with some other appropriate name and it does not extend `StorageManager`. So, this interface defines only about transactional behavior. 
 Any `StorageManager` which implements transactional behavior can implement this interface. `TransactionEventListener` can be instantiated with respective storagemanager or noop TransactionManager.

storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/provider/sql/factory/AbstractQueryExecutor.java
```diff
@@ -200,6 +195,63 @@ public void setStorableFactory(StorableFactory storableFactory) {
...
+                transactionBookKeeper.incrementNestedTransactionCount(Thread.currentThread().getId());
+            }
+        } catch (SQLException e) {
+            throw new StorageException("Failed to start transaction", e);
+        }
+    }
+
+    @Override
+    public void rollbackTransaction() throws StorageException {
+        Connection connection = transactionBookKeeper.getConnection(Thread.currentThread().getId());
```
satishd: You need to handle nested transactions here. Rollback should rollback the current connection effectively even when other nested transactions call commit.  
This can be done in the last end-transaction call which can be commit/rollback by maintaining effective state of the transaction in `TransactionBookKeeper`.

storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/provider/sql/factory/AbstractQueryExecutor.java
```diff
@@ -200,6 +195,63 @@ public void setStorableFactory(StorableFactory storableFactory) {
...
+            throw new StorageException("Failed to roll back transaction", e);
+        } finally {
+            closeConnection(connection);
+        }
+    }
+
+
+    @Override
+    public void commitTransaction() throws StorageException {
+        Connection connection = transactionBookKeeper.getConnection(Thread.currentThread().getId());
```
satishd: Need to handle nested transaction as mentioned in earlier comment in `rollback` method.

raju-saravanan:
@satishd: I have addressed your comments and tested my code changes, can you have another look?

storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/JdbcStorageManager.java
```diff
@@ -245,4 +243,18 @@ public void init(Map<String, Object> properties) {
         this.queryExecutor.setStorableFactory(storableFactory);
     }
 
+    @Override
+    public void beginTransaction(TransactionIsolation transactionIsolationLevel) throws StorageException {
```
satishd: These throw `StorageException` but `TransactionManager`'s methods doe snot throw any exceptions. We may not want to declare to throw any exception here. Same for rollback/commit methods.

raju-saravanan: Thanks Satish, I removed storage exception for QueryExecutor methods, looks like I missed for StorageManager methods. I will remove this.

satishd:
Thanks @raju-saravanan for your work.

PR #292, opened 2017-11-02
ISSUE-291 Remove findbugs-annotations and mariadb modules because of LGPL license.

satishd:
- Removed findbugs-annotation module dependencies
- Removed mariadb-client driver module is removed and respective driver class is replaced with mysql-connector
because of LGPL license issues.


satishd:
@raju-saravanan Can you verify this?

satishd:
@raju-saravanan Removed unused mariadb provided scope dependency in storage-tool. Pl verify running bootstrap with mysql instance.

PR #300, opened 2017-11-16
ISSUE-299 Close connections appropriately when the task for given connection is finished

satishd:
ISSUE-299 Close connections appropriately when the task for given connection is finished including both transactional and non-transactional semantics.


arunmahadevan:
+1

Issue #298, opened 2017-11-16
Default values are completely removed for Avro Records with Union fields


enoren:
This issue appears to have been introduced with Issue #88 which resolves issues with default values and union fields, however when a record has a field of union type all default values are removed and set to null in the resultant schema. For instance, I have a schema such as:

```
{
      "namespace": "avro",
      "type": "record",
      "name": "UnionRecord",
      "fields": [
          {
              "name": "union_field",
              "type": ["null", "string"],
              "default": null
          }, {
              "name": "int_field",
              "type": "int",
              "default": 5
          }
      ]
}
```

When pulling this in from the schema registry the field 'int_field' no longer has the default value so any migrations from old schemas are failing as a result. The default value is in the schema when looking at the schema registry UI as well as while tracing through the code it is pulled down and shows up in the schema up until the schema is processed through the com.hortonworks.registries.schemaregistry.avro.AvroSchemaResolver#handleUnionFieldsWithNull method at which point the returned schema has lost the default value for the "int_field".

The actual issue appears to be related to this line:

```
updatedFields.add(new Schema.Field(field.name(),
                                                   fieldSchema,
                                                   field.doc(),
                                                   currentFieldTypeIsUnion ? NullNode.getInstance() : null,
                                                   field.order()));
```

However if I change it to

```
updatedFields.add(new Schema.Field(field.name(),
                                                   fieldSchema,
                                                   field.doc(),
                                                   currentFieldTypeIsUnion ? NullNode.getInstance() : field.defaultValue(),
                                                   field.order()));
```

everything appears to work correctly again.

michaelandrepearce:
@enoren your logic change seems sensible for me, are you able to raise a PR?

enoren:
@michaelandrepearce Done - https://github.com/hortonworks/registry/pull/301

satishd:
Raised/merged #309 for adding tests

PR #301, opened 2017-11-16
Fix for Issue #298

enoren:
Not sure what branch you guys would want to put this on, but since we are working with 0.3.0 I based it on what I believe the commit is for that. As an aside, it appears that the v0.3.0 tag is incorrectly set on master instead of the 0.3.0 branch which I think is referenced in https://groups.google.com/forum/#!topic/registry/8GMrLlmhsic however maybe the tag fix was never pushed up?

### DIFF
--- a/bootstrap/sql/mysql/create_tables.sql
+++ b/bootstrap/sql/mysql/create_tables.sql
@@ -18,11 +18,11 @@
 
 CREATE TABLE IF NOT EXISTS schema_metadata_info (
   id              BIGINT AUTO_INCREMENT NOT NULL,
-  type            VARCHAR(256)          NOT NULL,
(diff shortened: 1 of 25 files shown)


michaelandrepearce:
@enoren pr should be to master

satishd:
@enoren,
No maint releases planned for 0.3.0, can you raise PR against master as @michaelandrepearce mentioned in earlier comment?

enoren:
Unfortunately it looks like changing the base didn't work for free so I will have to create a new PR based off of master when I get a chance.

Thanks

satishd:
@enoren Create a new branch forked from `master`, make the change you mentioned in #298 and raise PR with that branch.

enoren:
New PR https://github.com/hortonworks/registry/pull/303

PR #303, opened 2017-11-19
ISSUE-298 Pass-thru the default field values for non-union field types so default values are properly maintained and unaffected by the processing of union field type default values.

enoren:


satishd:
@enoren Thanks for contributing this PR, I will add a test for this another PR.

PR #307, opened 2017-11-27
ISSUE-306 Support default transaction isolation level

raju-saravanan:
ISSUE-306  Support default transaction isolation level


common/src/main/java/com/hortonworks/registries/common/transaction/TransactionIsolation.java
```diff
@@ -3,6 +3,7 @@
 import java.sql.Connection;
 
 public enum TransactionIsolation {
+    DEFAULT(-1),
```
satishd: nit: good to have documentation on this property.

PR #310, opened 2017-11-30
ISSUE-308: Registry UI Exception

shahsank3t:
Fixed various react libraries version since auto-updating (via npm install) breaks the UI.


satishd:
Thanks @shahsank3t for fixing this issue.

PR #321, opened 2017-12-07
fixed few UI libraries version

shahsank3t:


satishd:
@shahsank3t Did you face any issue with earlier versions?

shahsank3t:
@satishd I faced few issues in SAM and so for making it consistent for registry too, hence added a fix.

PR #377, opened 2018-01-11
ISSUE-378 Remove 401 page redirection as it can not handle non GET requests for now.

satishd:


webservice/src/main/resources/public/assets/401.html
```diff
@@ -27,7 +27,7 @@
     width: 1140px;
     margin: 0 auto;
     padding: 100px 0 50px;
-    background-image: url('back-default.png');
+    background-image: url('ui/back-default.png');
```
satishd: @shahsank3t Can you please check whether this change is right as I am not sure about html

arunmahadevan:
+1 After applying this patch, able to access registry in secure env.

priyank5485:
@satishd I had the same local changes. The only difference is i had the below for back-default.png

    background-image: 'url(\'../ui/styles/img/back-default.png\')';

However, I still ran in to the issue of favicon.ico GET request. Most likely its my local setup issue. As long as you and @arunmahadevan were able to see 401.html when you did not have a TGT in your cache and the actual resource when TGT was present we should be good.

arunmahadevan:
@priyank5485 , i only verified that registry UI was working after a kinit, did not check it without the ticket.

arunmahadevan:
the env is deleted so I cannot validate it now.

satishd:
@arunmahadevan  @priyank5485 I validated with and without kinit and both the pages were comingup fine.
screen shot 2018-01-11 at 5 48 48 pm

satishd:
Closing this PR as removing error page redirection will be handled in separate PR.

PR #383, opened 2018-01-12
(WIP) ISSUE-381 Provide helper utility for manual transaction handling

### LINKED ISSUE #381: Provide helper utility for manual transaction handling
Recently Streamline encountered a blocker issue (https://github.com/hortonworks/streamline/issues/1157) which came from very silly mistake (patch: https://github.com/hortonworks/streamline/pull/1158).

This is the logic for handling with manual transaction (after applying fix):

```
    @FunctionalInterface
    interface ConsumerWithThrowsException<T> {
        void accept(T t) throws Exception;
    }

    private void mutateTopologyContext(ConsumerWithThrowsException<TopologyContext> mutatingFunction,
                                       TopologyContext topologyContext) throws Exception {
        boolean rolledBackTransaction = false;
        try {
            transactionManager.beginTransaction(TransactionIsolation.DEFAULT);
            mutatingFunction.accept(topologyContext);
        } catch (IgnoreTransactionRollbackException e) {
            throw convertThrowableToException(e.getCause());
        } catch (Throwable e) {
            transactionManager.rollbackTransaction();
            rolledBackTransaction = true;
            throw convertThrowableToException(e);
        } finally {
            if (!rolledBackTransaction) {
                transactionManager.commitTransaction();
            }
        }
    }

    private void mutateTopologyContext(ConsumerWithThrowsException<TopologyContext> mutatingFunction,
                                       Topology topology, String asUser) throws Exception {
        mutateTopologyContext(mutatingFunction, getTopologyContext(topology, asUser));
    }

```

which IgnoreTransactionRollbackException acts as a marker exception to guide that we should commit transaction instead of rolling back even exception occurs.

https://github.com/HeartSaVioR/streamline/blob/2a8c3ecb930f88211f20598d104a33aa9e8c88bd/common/src/main/java/com/hortonworks/streamline/common/exception/IgnoreTransactionRollbackException.java

We could generalize above code block and provide utility class which contains methods accepting some functional interfaces (like Consumer, Function, etc.) and handling manual transaction. It should help avoiding silly mistakes which affect really badly.

@satishd Do you think this issue is making sense to go with crafting and submitting a patch?

HeartSaVioR:
* add methods to TransactionManager which helps users to avoid silly mistakes
* describe method contract in methods

This closes #381


HeartSaVioR:
Let's continue discussion based on the code. I guess I've put inputs into the PR, but the PR still needs to be reviewed and something can be changed, hence I didn't address javadoc. 
Will address javadoc for them when we finalize method signatures, and its place.

storage/core/src/main/java/com/hortonworks/registries/storage/TransactionManager.java
```diff
@@ -27,13 +28,87 @@
...
+     *
+     * Unlike of rollbackTransaction(), implementation of this method doesn't need to guarantee
+     * transaction is closed, since in normal caller would want to call rollbackTransaction()
+     * when commitTransaction() throws Exception.
      */
     void commitTransaction();
+
+    // TODO: javadoc
+    default <T, R> R runInTransaction(TransactionIsolation transactionIsolation,
+                                      TransactionManagerAwareFunctionCapableOfThrowingException<T, R> fn, T t) throws Exception {
```
arunmahadevan: Why should the function be aware of TransactionManager? 

The method is already invoked on a TransactionManager object, so why not just take a FunctionThrowingException ?

arunmahadevan: Whats the third argument "t" ? Shouldn't we accept a some block of code here (like a no-arg function that returns nothing) ?

storage/core/src/main/java/com/hortonworks/registries/storage/TransactionManager.java
```diff
@@ -27,13 +28,87 @@
...
+        final TransactionManager tm = this;
+        return runInTransactionInternal(transactionIsolation, (_void) -> fn.apply(tm, t));
+    }
+
+    // TODO: javadoc
+    default <T> void runInTransaction(TransactionIsolation transactionIsolation,
+                                      TransactionManagerAwareConsumerCapableOfThrowingException<T> fn, T t) throws Exception {
```
arunmahadevan: Why should the function be aware of TransactionManager?

storage/core/src/main/java/com/hortonworks/registries/storage/TransactionManager.java
```diff
@@ -27,13 +28,87 @@
...
+        final TransactionManager tm = this;
+        runInTransactionInternal(transactionIsolation, (_void) -> {
+            fn.accept(tm, t);
+            return null;
+        });
+    }
+
+    default <R> R runInTransactionInternal(TransactionIsolation transactionIsolation,
```
arunmahadevan: since you define in interface, this will be part of the public API, so defeats the purpose of internal.

storage/core/src/main/java/com/hortonworks/registries/storage/TransactionManager.java
```diff
@@ -27,13 +28,87 @@
...
+            committed = true;
+            throw convertThrowableToException(e.getCause());
+        } finally {
+            if (!committed) {
+                rollbackTransaction();
+            }
+        }
+    }
+
+    default Exception convertThrowableToException(Throwable t) {
```
arunmahadevan: since you define in interface, this will be part of the public API

storage/core/src/main/java/com/hortonworks/registries/storage/TransactionManager.java
```diff
@@ -27,13 +28,87 @@
...
+     * Flushes the changes made to the storage layer.
+     *
+     * Unlike of rollbackTransaction(), implementation of this method doesn't need to guarantee
+     * transaction is closed, since in normal caller would want to call rollbackTransaction()
+     * when commitTransaction() throws Exception.
      */
     void commitTransaction();
+
+    // TODO: javadoc
+    default <T, R> R runInTransaction(TransactionIsolation transactionIsolation,
```
arunmahadevan: Maybe define another method where the user does not need to pass the isolation level and we use the default.

storage/core/src/main/java/com/hortonworks/registries/storage/TransactionManager.java
arunmahadevan: nit: could use wildcard (?) instead of R here.

HeartSaVioR:
@arunmahadevan 
My idea was based on creating utility class, and @satishd suggested adding methods to interface. I guess I messed up with adding methods which are designed to utility class to interface.

I'd like to hear our voices regarding which is convenient: adding utility methods to interface (and having utility class internally which will contain refactored internal logic) vs adding utility class for public.

HeartSaVioR:
I don't mind whoever submits another patch which API looks nice and well documented. So if someone has well defined thought regarding API spec and documentation, please feel free to go ahead with another patch.

storage/core/src/main/java/com/hortonworks/registries/storage/TransactionManager.java
```diff
@@ -27,13 +28,87 @@
...
+            throw (Error) t;
+        } else if (t instanceof Exception) {
+            return (Exception) t;
+        } else {
+            return new RuntimeException(t);
+        }
+    }
+
+    @FunctionalInterface
+    interface FunctionCapableOfThrowingException<T, R> {
```
raju-saravanan: FunctionCapableOfThrowingException takes one input and produces one output, I have seen scenarios in SAM where transaction block will take more than one input, any idea how should we go about handling such cases?

raju-saravanan:
@HeartSaVioR: I think moving your methods to a Util class would be more preferable.

satishd:
There are different ways to give this functionality with respective APIs either through TransactionManager API or some utils API. You can either take Runnable or FunctionalInterface which throws Exception for that API for running the given code in a transaction. It is good to refine those thoughts before putting into code. I am fine with any approach having a good rationale of the choice made.

HeartSaVioR:
Just revised the patch.

ManagedTransaction class is responsible to execute transaction code block (supporting Function and Consumer) with managed transaction. I created 0~5 args of Function and Consumer, hopefully it is enough for us to use. We can add more, but someone should consider entity for parameters instead of too-many-parameters.

@arunmahadevan @satishd @raju-saravanan 
Please review this again. Thanks in advance.

storage/core/src/main/java/com/hortonworks/registries/storage/transaction/ManagedTransaction.java
```diff
@@ -0,0 +1,272 @@
...
+        if (t instanceof Error) {
+            throw (Error) t;
+        } else if (t instanceof Exception) {
+            return (Exception) t;
+        } else {
+            return new RuntimeException(t);
+        }
+    }
+
+    private interface FunctionCapableOfThrowingException<T, R> {
```
arunmahadevan: nit: this can be a `SupplierCapableOfThrowingException<R>` since the only usage does not take any argument.

HeartSaVioR: Nice catch. Addressed.

arunmahadevan:
+1

storage/core/src/test/java/com/hortonworks/registries/storage/transaction/ManagedTransactionTestHelper.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public void throwException(Object...args) throws Exception {
+        call(args);
+        throw INTENDED_EXCEPTION;
+    }
+
+    public void throwIgnoreRollbackException(Object...args) throws Exception {
+        call(args);
+        throw IGNORE_ROLLBACK_EXCEPTION;
```
satishd: nit: better to throw desired Exception instead of using an instance which fills the stacktrace when this instance is created. I understand that stacktrace is not used in assertion but it may be  good to avoid using a created Exception.

HeartSaVioR: Addressed. I don't replace INTENDED_EXCEPTION since it is used as comparison. (I guess you are aware of this, but just be clear. :) )

satishd: I am aware of that but I prefer to create a specific Exception class and check for that class instead of an `Exception` instance.

HeartSaVioR: Makes sense. Will address too.

HeartSaVioR: Address. Please take a look again at the latest commit. (Note: commit amended)

satishd:
Thanks @HeartSaVioR for addressing review comments.

PR #403, opened 2018-02-08
ISSUE-402 Registry 'Jar Storage HDFS URL' does not work for Namenode HA

snemuri:
This fix introduces a new variable called $HADOOP_CONF_DIR which can be used to configure the location of core-site.xml and hdfs-site.xml, which will then be added to registry classpath during startup. After this patch, Add "export HADOOP_CONF_DIR=/etc/hadoop/conf" in registry-env to configure NN HA.


arunmahadevan:
+1

arunmahadevan:
@satishd , please take a look and merge.

snemuri:
@satishd Could you please review this ?

PR #405, opened 2018-02-09
ISSUE-404: Support http proxy related configuration for registry server

priyank5485:
Support http proxy related configuration for registry server

### DIFF
--- a/conf/registry-inmemory-example.yaml
+++ b/conf/registry-inmemory-example.yaml
@@ -60,11 +60,6 @@ swagger:
 #enable CORS, may want to disable in production
 enableCors: false
 
-#Set below 3 properties if server needs a proxy to connect. Useful to download mysql jar
-#httpProxyUrl: "http://proxyHost:port"
-#httpProxyUsername: "username"
-#httpProxyPassword: "password"
-
 server:
   applicationConnectors:
     - type: http
--- a/conf/registry-inmemory-example.yaml
+++ b/conf/registry-inmemory-example.yaml
@@ -60,6 +60,11 @@ swagger:
 #enable CORS, may want to disable in production
 enableCors: false
 
+#Set below 3 properties if server needs a proxy to connect. Useful to download mysql jar
+#httpProxyUrl: "http://proxyHost:port"
+#httpProxyUsername: "username"
+#httpProxyPassword: "password"
+
 server:
   applicationConnectors:
     - type: http
--- a/conf/registry-oracle-example.yaml
+++ b/conf/registry-oracle-example.yaml
@@ -67,6 +67,11 @@ swagger:
 #enable CORS, may want to disable in production
 enableCors: false
 
+#Set below 3 properties if server needs a proxy to connect. Useful to download mysql jar
+#httpProxyUrl: "http://proxyHost:port"
+#httpProxyUsername: "username"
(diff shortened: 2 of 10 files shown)


priyank5485:
@arunmahadevan That is correct. Yes, ultimately the goal is to have only one set of configs at top level. For that, we need a change in streamline code to pick those configs up from top level. This change in registry seemed independent and was to fix the proxy issue related to downloading mysql jar in TablesInitializer call, that was being used in streamline and registry bootstrap-storage.sh call. I think we can independently merge this and raise a separate PR to handle streamline change to avoid that config at two levels. Let me know if you think otherwise

priyank5485:
I have also added those 3 fields to RegistryConfiguration and squashed and pushed my branch.

sapinamin:
I have submitted the ambari PR https://github.com/hortonworks/hdf_ambari_mp/pull/187 and tested it with jar with this PR. Please merge it.

arunmahadevan:
Are the changes proposed in this PR still required. Does https://github.com/hortonworks/streamline/pull/1215 override this ?

arunmahadevan:
@priyank5485 , ignore the previous comment. I didnt realize this patch is for registry and the other one is for streamline.

So the question is: does https://github.com/hortonworks/streamline/pull/1215 depend on the changes proposed in this patch?

conf/registry-oracle-example.yaml
```diff
@@ -67,6 +67,11 @@ swagger:
 #enable CORS, may want to disable in production
 enableCors: false
 
+#Set below 3 properties if server needs a proxy to connect. Useful to download mysql jar
```
arunmahadevan: nit : i dont think we should reference mysql in registry-oracle-example.yaml

conf/registry-inmemory-example.yaml
```diff
@@ -60,6 +60,11 @@ swagger:
 #enable CORS, may want to disable in production
 enableCors: false
 
+#Set below 3 properties if server needs a proxy to connect. Useful to download mysql jar
```
arunmahadevan: nit : i dont think we should reference mysql in this file

conf/registry-postgres-example.yaml
```diff
@@ -67,6 +67,11 @@ swagger:
 #enable CORS, may want to disable in production
 enableCors: false
 
+#Set below 3 properties if server needs a proxy to connect. Useful to download mysql jar
```
arunmahadevan: nit : i dont think we should reference mysql in this file

conf/registry-oracle-example.yaml
priyank5485: Addressed

conf/registry-inmemory-example.yaml
priyank5485: Addressed

conf/registry-postgres-example.yaml
priyank5485: Addressed

priyank5485:
@arunmahadevan I have addressed your comments and removed mysql references in non mysql yaml files. Streamline PR that you asked about depends on this PR for downloading mysql jar during setup.

arunmahadevan:
+1

PR #419, opened 2018-03-15
ISSUE-418: Add JWT based SSO as alternative authentication to common-auth

priyank5485:
ISSUE-418: Add JWT based SSO as alternative authentication to common-auth module.

Please note that most of the code in this PR has been copied from hadoop project at https://github.com/apache/hadoop/tree/branch-3.0.0 with the corresponding license

Other projects relying on registry project's common-auth module can leverage code from this PR to implement SSO behavior using the User-Agent request header

### DIFF
--- a/common-auth/pom.xml
+++ b/common-auth/pom.xml
@@ -43,9 +43,21 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>com.nimbusds</groupId>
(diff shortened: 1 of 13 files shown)


priyank5485:
I have tested this PR with a knox jwt based sso using a local ldap setup provided by knox

PR #437, opened 2018-04-18
ISSUE-436 Respect contract described in StorageManager for JdbcStorageManager

### LINKED ISSUE #436: JdbcStorageManager doesn't respect the contract in StorageManager
The contract in StorageManager is a bit tough since it is designed based on in-memory first which is easy to achieve this without worrying about overhead.

I found that JdbcStorageManager doesn't respect the contract in StorageManager: some operation is already respecting it but without transactional manner. It may hurt overall performance on usage if we make JdbcStorageManager implementing StorageManager correctly, but looks like both registry and streamline are relying on such contract, so might be better to ensure once, and look for alternative approaches.

HeartSaVioR:
* ensure that JdbcStorageManager implements StorageManager correctly with transactional manner
* this assumes we're aware about the impact on performance

This closes #436

### DIFF
--- a/storage/core/src/main/java/com/hortonworks/registries/storage/exception/NotFoundException.java
+++ b/storage/core/src/main/java/com/hortonworks/registries/storage/exception/NotFoundException.java
@@ -1,8 +1,0 @@
-package com.hortonworks.registries.storage.exception;
-
-public class NotFoundException extends StorageException {
-
-    public NotFoundException(String message) {
-        super(message);
-    }
-}
--- a/storage/core/src/main/java/com/hortonworks/registries/storage/exception/NotFoundException.java
+++ b/storage/core/src/main/java/com/hortonworks/registries/storage/exception/NotFoundException.java
@@ -1,0 +1,8 @@
+package com.hortonworks.registries.storage.exception;
+
+public class NotFoundException extends StorageException {
+
+    public NotFoundException(String message) {
+        super(message);
+    }
+}
--- a/storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/JdbcStorageManager.java
+++ b/storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/JdbcStorageManager.java
@@ -28,7 +28,7 @@ import com.hortonworks.registries.storage.StorageManager;
 import com.hortonworks.registries.storage.TransactionManager;
 import com.hortonworks.registries.storage.exception.AlreadyExistsException;
 import com.hortonworks.registries.storage.exception.IllegalQueryParameterException;
-import com.hortonworks.registries.storage.exception.NotFoundException;
+import com.hortonworks.registries.storage.exception.StorableNotFoundException;
 import com.hortonworks.registries.storage.exception.StorageException;
 import com.hortonworks.registries.storage.impl.jdbc.provider.QueryExecutorFactory;
 import com.hortonworks.registries.storage.impl.jdbc.provider.sql.factory.QueryExecutor;
@@ -141,7 +141,7 @@ public class JdbcStorageManager implements TransactionManager, StorageManager {
 
             final Storable existing = get(storable.getStorableKey());
             if (existing == null) {
-                throw new NotFoundException("Entity not found for such key: " + storable.getStorableKey());
+                throw new StorableNotFoundException("Entity not found for such key: " + storable.getStorableKey());
             }
             queryExecutor.update(storable);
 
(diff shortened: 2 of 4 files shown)


storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/JdbcStorageManager.java
```diff
@@ -61,29 +62,96 @@ public JdbcStorageManager(QueryExecutor queryExecutor) {
 
     @Override
     public void add(Storable storable) throws AlreadyExistsException {
-        log.debug("Adding storable [{}]", storable);
-        queryExecutor.insert(storable);
+        boolean committed = false;
+        try {
+            beginTransaction(TransactionIsolation.DEFAULT);
```
arunmahadevan: the transactions are handled at a higher level via [ManagedTransaction](https://github.com/hortonworks/registry/blob/master/storage/core/src/main/java/com/hortonworks/registries/storage/transaction/ManagedTransaction.java).

I dont think the Storage Manager should support txn for individual operations. There could be multiple storage manager operations that needs to be part of a txn.

HeartSaVioR:
@arunmahadevan 
Nested transaction is supported now, so that's not impossible to use transaction in each operation. @raju-saravanan Could you confirm?
And there's currently no way to get current transaction, so can't conditionally wrap the logic with transaction. The implementation of JdbcStorageManager should not depend on transaction outside, and such requirement should be documented if it should.

Now the storage module is a common library for three projects, and this is not the first time I claim about broken contract. so I would want to do the thing the right way.

cc. @satishd

raju-saravanan:
@HeartSaVioR and @arunmahadevan Yes, we support nested transactions. If a transaction is nested then it uses the connection details of the outer transaction, it commits or rollbacks with the outermost transaction block.

storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/JdbcStorageManager.java
```diff
@@ -61,29 +62,96 @@ public JdbcStorageManager(QueryExecutor queryExecutor) {
...
     public <T extends Storable> T remove(StorableKey key) throws StorageException {
-        T oldVal = get(key);
-        if (key != null) {
-            log.debug("Removing storable key [{}]", key);
-            queryExecutor.delete(key);
+        boolean committed = false;
+        try {
+            beginTransaction(TransactionIsolation.DEFAULT);
+
+            T oldVal = get(key);
```
raju-saravanan: get() here may not be required. As by default "delete" command with "where" clause which doesn't match any records when executed doesn't throw any error, but quietly prints that "0 rows deleted". I confirmed this behavior in MySQL, Postgres and Oracle.

storage/core/src/main/java/com/hortonworks/registries/storage/exception/NotFoundException.java
```diff
@@ -0,0 +1,8 @@
+package com.hortonworks.registries.storage.exception;
+
+public class NotFoundException extends StorageException {
```
raju-saravanan: nit: There are couple of "[a-zA-Z]*NotFoundException", maybe we can call this "StorableNotFoundException" to distinguish them from this exception :)

HeartSaVioR: Nice finding. Will fix.

storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/JdbcStorageManager.java
HeartSaVioR: The method needs to return old value as method contract / signature, so need to get old value first.

raju-saravanan: I see that there are instances were storage entity obtained by calling remove() is being used and certain places where we just discard the value. I think we should have remove() not return any values and if the user wants to fetch the entity which is being removed then he can do get() and followed by remove(). I know this is an existing behavior not related to this PR, if you agree with the proposal we can come back to it in a different PR.

HeartSaVioR:
Thanks @raju-saravanan for the reviewing. I'll wait for @arunmahadevan and @satishd to review this, and merge once either one approves the patch.

HeartSaVioR:
@satishd @arunmahadevan  
As I said earlier, I'm not sure it is safe to rely on Postgre's error code, so don't want to do it on my own. If anyone would like to do it, please feel free to take the issue up.

If we can't make sure the implementation respects the interface, I would like to keep others as it is, and modify only addOrUpdate to do get -> add/update guarding with transaction. It would be easier to make a change to JdbcStorageManager, but I'm OK to give a try to make a change in PostgresqlExecutor. Does it make sense then?

satishd:
@HeartSaVioR
>but I'm OK to give a try to make a change in PostgresqlExecutor. Does it make sense then?
I agree on making a change in `PostgresqlExecutor#insertOrUpdate`.

HeartSaVioR:
>> but I'm OK to give a try to make a change in PostgresqlExecutor. Does it make sense then?

> I agree on making a change in PostgresqlExecutor#insertOrUpdate.

OK I can handle that. Will address from #441.

Issue #438, opened 2018-04-19
UI hangs if schemaBranches is empty


oscarandreu:
If a given schema have it branches empty the UI is unable to load.
The problem is that the schemaBranches property is being used without any previous check giving the next error:
TypeError: Cannot read property 'schemaVersionInfos' of undefined
    at new t (application.js:23)

To reproduce the error only need a clean HDF sandbox and execute this:
`curl -X DELETE 'http://127.0.0.1:7788/api/v1/schemaregistry/schemas/trucking_data_traffic:v/versions/1'`

oscarandreu:
I did it directly from the database.

gcsaba2:
Closed due to inactivity.

PR #439, opened 2018-04-19
UI hangs if schemaBranches is empty #438

oscarandreu:
UI hangs if schemaBranches is empty #438

From my point of view the change it's not complete, still remains to know what happens in case there is no version.


oscarandreu:
Old one

PR #443, opened 2018-04-26
MINOR: Docker db hostnames are exposed to the user.

kamalcph:
- To connect the remote / external db instance, user might want to know the IP of the db, so exposed it.
- JCE doc updated.


kamalcph:
@satishd @raju-saravanan 

Could you please review this PR ?

PR #445, opened 2018-05-15
JDBC HA Leadership Participant Implementation

joanvr:
Although it really makes sense to use ZK to coordinate HA leadership, taking into consideration we already need it for Kafka, sometimes we may not have access to ZK, or we may have a shared registry across different environments and it doesn't make sense to tie it to any specific one.
Taking into consideration that we already have a Storage dependency on MySQL or PostgreSQL, it makes sense to also coordinate HA leadership through it.
This implementation uses a table to manage the leader election. Taking advantage of primary key collision, we can handle only one leader at the same time.
More details on the code comments


michaelandrepearce:
Why is this needed if you are already using a db?

When using a DB you can essentially have multi master service as the db backend has the transactionality. That's there today

PR #451, opened 2018-06-07
fix download of apacheds-jdbm1 jar instead of bundle

vion1:
apacheds-jdbm1-2.0.0-M2.bundle is no longer available on nexus so the build will fail.   This fix will download the correct artifact apacheds-jdbm1-2.0.0-M2.jar from nexus.

### DIFF
--- a/common-auth/pom.xml
+++ b/common-auth/pom.xml
@@ -103,6 +103,14 @@
         </resource>
       </resources>
         <plugins>
+        <!-- apacheds-jdbm1-2.0.0-M2.bundle is no longer available on nexus.
+             This fix will download the correct artifact apacheds-jdbm1-2.0.0-M2.jar from nexus. -->
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <inherited>true</inherited>
+                <extensions>true</extensions>
+            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>


satishd:
[~von] I built using `mvn -B -nsu -B -nsu clean install -DskipTests -Pdist` on HDF - 3.1-maint and HDF - 3.1.1.16 successfully on my mac. 

maven version used is:
{code:java}
mvn --version
Apache Maven 3.5.3 (3383c37e1f9e9b3bc3df5050c29c8aff9f295297; 2018-02-25T01:19:05+05:30)
Maven home: /usr/local/Cellar/maven/3.5.3/libexec
Java version: 1.8.0_161, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_161.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.13.4", arch: "x86_64", family: "mac"
{code}

It may be good to find the root cause on why this is failing on specific machines instead of adding this change here.

common-auth/pom.xml
```diff
@@ -103,6 +103,12 @@
         </resource>
       </resources>
         <plugins>
+            <plugin>
```
satishd: @vion1 Can you put a comment with details on why this was added so that others would know the reason. As you mentioned in the bug, this is because of apacheds-jdbm1-2.0.0-M2.bundle not available now.

vion1: apacheds-jdbm1-2.0.0-M2.bundle is no longer available on nexus so the build will fail.   This fix will download the correct artifact apacheds-jdbm1-2.0.0-M2.jar from nexus.

satishd: @vion1 Can you put the above comment in the file at that line(which is 106) so that whoever is looking at the file can know the reason?

PR #452, opened 2018-06-19
Updated lib versions and resolved dependency conflicts

satishd:
Updated different library versions to address CVEs.


satishd:
@raju-saravanan Can you review this change and also verify whether ser/des for avro works fine when it is run with schema-registry-serdes modules?

satishd:
@arunmahadevan I am not sure whether streamline will have any issues as we few libraries are updated here. You may want to try out these changes in streamline environment and make appropriate change if required.

PR #453, opened 2018-06-20
MINOR Added APIs to build schema registry version

satishd:


schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/SchemaRegistryResource.java
```diff
@@ -101,9 +102,22 @@
...
+                                  SchemaRegistryVersion schemaRegistryVersion) {
         super(schemaRegistry, leadershipParticipant);
+        this.schemaRegistryVersion = schemaRegistryVersion;
+    }
+
+    @GET
+    @Path("/version")
+    @ApiOperation(value = "Get version info of this Schema Registry instance",
+            response = SchemaRegistryVersion.class)
+    @Timed
```
raju-saravanan: nit: Can we add a tag here, so that this API would show under appropriate category in swagger?

satishd:
@raju-saravanan Added version related classes into common-client module so that clients cal also get to know which library version are they using. Can you review the latest commit?

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
```diff
@@ -996,6 +1000,10 @@ public void close() {
         client.close();
     }
 
+    public SchemaRegistryVersion clientVersion() {
+        return CLIENT_VERSION;
+    }
+
     private <T> T createInstance(SerDesInfo serDesInfo, boolean isSerializer) {
```
raju-saravanan: Should we expose a function to get the SchemaRegistryVersion of the server to which the client will be interacting with ?

satishd: @raju-saravanan I thought about that while adding `clientVersion` API to add a new API like `serverVersion` with default method in `ISchemaRegistryService` but left that as I did not see any need for now.

raju-saravanan: ok

PR #466, opened 2018-07-28
MINOR : Fix flyway shell migration error

raju-saravanan:


raju-saravanan:
@arunmahadevan: It was failing for migrations involving shell scripts, for SQLs it was working fine.

PR #468, opened 2018-08-01
ISSUE-467 UI fails to load schemas without schema versions

shahsank3t:
If no schema versions, the user cannot expand the panel nor add another version from the UI.
Either add version directly through API's from the backend or create a new schema altogether.

@satishd - Please let me know if this behavior works in general since from UI there's rare chance that the user might fall into this issue or would you prefer that we should revisit the flow for such cases.


satishd:
@shahsank3t This behavior needs to be changed. If there is schema metadata with out any versions then we should allow user to create schema version.

shahsank3t:
@satishd - updated the flow. The following screen will be visible to the user when no schema versions are found and the user can click on edit (pencil icon) to add new version which will by default be added into MASTER branch.

screen shot 2018-08-02 at 1 54 38 pm

satishd:
@raju-saravanan You can raise github issue for delete support in master branch which can be looked into later.

PR #469, opened 2018-08-01
ISSUE-465 replace ClientBuilder with JerseyClientBuilder

ngorchakova:
replace ClientBuilder with JerseyClientBuilder to avoid collisions in case of multiple implementations


michaelandrepearce:
Looks good to me, nice find and simple fix.

PR #474, opened 2018-08-13
ISSUE-485 Fix adding large schema text for oracle

raju-saravanan:
Fix ORA-01461 (https://support.oracle.com/knowledge/Middleware/1620109_1.html) when using the "merge" command which is used in addOrUpdate. addOrUpdate is used while updating schemaVersionStorable's state after persisting schemaStateStorable, so instead of using addOrUpdate use update as it serves the same purpose.

### DIFF
--- a/registry-dist/VERSION
+++ b/registry-dist/VERSION
@@ -1,0 +1,6 @@
+#Created by build system. Do not modify
+#Mon Aug 13 23:06:05 IST 2018
+version=0.5.1-SNAPSHOT
+revision=cf2c6e1cf0021d4440ceb53a46a6c3910907bba4
+name=Registry Binary Distribution
+timestamp=1534181765088
--- a/registry-dist/VERSION
+++ b/registry-dist/VERSION
@@ -1,6 +1,0 @@
-#Created by build system. Do not modify
-#Mon Aug 13 23:06:05 IST 2018
-version=0.5.1-SNAPSHOT
-revision=cf2c6e1cf0021d4440ceb53a46a6c3910907bba4
-name=Registry Binary Distribution
-timestamp=1534181765088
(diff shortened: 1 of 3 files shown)


registry-dist/VERSION
```diff
@@ -0,0 +1,6 @@
+#Created by build system. Do not modify
```
satishd: You can remove this file, it may have been added accidentally into this commit.

schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/SchemaVersionLifecycleManager.java
```diff
@@ -847,7 +847,7 @@ private void storeSchemaVersionState(SchemaVersionLifecycleContext schemaVersion
             throw new SchemaNotFoundException("No Schema version exists with id " + schemaVersionId);
         }
         versionedSchema.setState(stateId);
-        storageManager.addOrUpdate(versionedSchema);
+        storageManager.update(versionedSchema);
```
satishd: Right, it seems addOrUpdate may not be needed here as it already exists.

storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/provider/oracle/query/OracleUpdateQuery.java
```diff
@@ -62,8 +63,10 @@ protected String createParameterizedSql() {
     }
 
     private Map<Schema.Field, Object> createWhereClauseColumnToValueMap() {
-        Map<Schema.Field, Object> bindingMap = getBindings().stream().
-                collect(Collectors.toMap(keyValuePair -> keyValuePair.getKey(), keyValuePair -> keyValuePair.getValue()));
+        Map<Schema.Field, Object> bindingMap = new HashMap<>();
```
satishd: @raju-saravanan confirmed offline that earlier `stream().collect(Collectors.toMap)` operation throws an error if there are duplicate keys.

satishd:
https://github.com/hortonworks/registry/commit/559455a2f212416ddedae0b70e972edcec5a7566 is merged into HDF-3.1-maint

PR #484, opened 2018-08-23
ISSUE-486 Fix deadlocks in storage layer during concurrent calls to SR

raju-saravanan:
In case of code path for adding schema metadata, we first do a get(schemaMetadataName), if the get returns a null then we add the new schemaMetadata else we return the schemaMetadataId from the get() call. The current transaction isolation level in SR is SERIALIZABLE, in this case, the get() call gets a shared lock and the add() call get an exclusive lock within a transaction. If two threads are invoking this API we might run into a situation where thread1 has obtained a shared lock and thread2 has also obtained the shared lock but thread1 can't get an exclusive lock because of a shared lock held by thread2 and thread2 can't get an exclusive lock because of a shared lock held by thread1. So, we run into a deadlock. The solution here would be to reduce the transaction isolation level to REPETABLE_READS for MySQL and POSTGRES and since ORACLE doesn't use this locking mechanism to in case of SERIALIZABLE, we can retain the current transaction isolation level for ORACLE.


conf/registry.yaml
```diff
@@ -64,15 +64,15 @@ fileStorageConfiguration:
...
- providerClass: "com.hortonworks.registries.storage.impl.jdbc.JdbcStorageManager"
- properties:
-   db.type: "mysql"
-   queryTimeoutInSecs: 30
-   db.properties:
-     dataSourceClassName: "com.mysql.jdbc.jdbc2.optional.MysqlDataSource"
-     dataSource.url: "jdbc:mysql://localhost/schema_registry"
-     dataSource.user: "registry_user"
-     dataSource.password: "registry_password"
+  providerClass: "com.hortonworks.registries.storage.impl.jdbc.JdbcStorageManager"
```
priyank5485: Any reason for changing mysql to postrgre? Comment still says MySQL.

registry-dist/VERSION
```diff
@@ -0,0 +1,6 @@
+#Created by build system. Do not modify
```
priyank5485: Remove this file?

schema-registry/rest-service/pom.xml
```diff
@@ -44,6 +44,18 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <!-- https://mvnrepository.com/artifact/mysql/mysql-connector-java -->
+      <dependency>
```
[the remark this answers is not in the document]
satishd: Right, this dependency should be removed. There was a licensing issue to package with registry distribution. bootstrap-storage.sh downloads the required jar and puts into respective lib dir.

raju-saravanan:
@satishd  and @priyank5485: This PR is a WIP. I simply created this so that one of the break fix engineers could test this further, the PR in itself is not complete.

satishd:
>@satishd and @priyank5485: This PR is a WIP. I simply created this so that one of the break fix engineers could test this further, the PR in itself is not complete.

@raju-saravanan Sure, I will hold my review till you confirm. 👍

raju-saravanan:
@satishd  and @priyank5485: Can you guys review this PR?

satishd:
@raju-saravanan Can you rebase/resolve the conflicts?

raju-saravanan:
@satishd : Rebased

storage/core/src/main/java/com/hortonworks/registries/storage/transaction/TransactionEventListener.java
```diff
@@ -90,18 +90,12 @@ public void onEvent(RequestEvent event) {
                     isTransactionActive = true;
                 }
             } else if (eventType == RequestEvent.Type.RESP_FILTERS_START) {
-                // not supporting transactions to filters
```
satishd: We may soon forget why this change was done. :) 
You can put a comment on why this change is done.

raju-saravanan: @satishd  Done

priyank5485:
+1

PR #491, opened 2018-09-18
ISSUE-489: Add docs on supporting Kafka Message Header to save schemaVersionId

kamalcph:


docs/serdes.rst
```diff
@@ -2,17 +2,41 @@
...
+Message Formats
+```````````````
+
+In Kafka v0.11.0.0, Record Header support is introduced. With this support, user can optionally save the schema
+information in the header rather than prepending it in the body of the message. (See Message Format 2)
+
+Message Format 1
+""""""""""""""""
+|header:| |body:<version_info><payload>|
+Prepends the schema version information along with the user payload in the body of the Message.
```
satishd: We may want to add this format is the default format to maintain backward compatibility. It is good to mention how users can use message format 2. Incase of Kafka serializer/deserializer, add details on how to use headers based schema version serialization [here]( https://github.com/hortonworks/registry/blob/master/docs/schema-registry.rst#kafka-producer-integration-with-schemaregistry)

kamalcph: Addressed your comments. Please take another look.

PR #501, opened 2018-10-10
ISSUE-500: Changes related to enabling JWT based SSO along with SPNEG…

priyank5485:
…O/Kerberos

### DIFF
--- a/common-auth/src/main/java/com/hortonworks/registries/auth/server/JWTAuthenticationHandler.java
+++ b/common-auth/src/main/java/com/hortonworks/registries/auth/server/JWTAuthenticationHandler.java
@@ -23,6 +23,7 @@ import javax.servlet.http.HttpServletResponse;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
+import java.util.Locale;
 import java.util.Properties;
 import java.text.ParseException;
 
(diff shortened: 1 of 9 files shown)


priyank5485:
@omkreddy I have tested this on a kerberized and sso enabled cluster and it works. Can you please review it?

common-auth/src/main/java/com/hortonworks/registries/auth/server/JWTAuthenticationHandler.java
```diff
@@ -368,4 +387,33 @@ protected boolean validateExpiration(SignedJWT jwtToken) {
...
+     * A User-Agent String is considered to be a browser if it does not contain
+     * any of the values from non-browser.user-agents; the default
+     * behavior is to consider everything a browser unless it contains one of:
+     * "java", "curl", "wget", "perl" or "Jersey".  Subclasses can optionally override
+     * this method to use different behavior.
+     *
+     * @param userAgent The User-Agent String, or null if there isn't one
+     * @return true if the User-Agent String refers to a browser, false if not
+     */
+    private boolean isBrowser(String userAgent) {
```
omkreddy: Can we avoid duplicating this method between handler classes? maybe we can move to util class?

common-auth/src/main/java/com/hortonworks/registries/auth/server/JWTAuthenticationHandler.java
```diff
@@ -147,6 +149,14 @@ public void init(Properties config) throws ServletException {
         if (customCookieName != null) {
             cookieName = customCookieName;
         }
+
+        nonBrowserUserAgents = config.getProperty(
```
omkreddy: Can we avoid duplicating code between handler classes? maybe we can create  getNonBrowserUserAgents() util method?

priyank5485:
@omkreddy Thanks for the review. I have addressed your review comments.

PR #503, opened 2018-10-10
ISSUE-502: Field props stripped in deserialization

### LINKED ISSUE #502: Field props stripped in deserialization
During deserialization that passes through `AvroSchemaResolver`, field level props are stripped from the schema on the resultant object. This only occurs if any field in the schema includes a union type; I've narrowed it down to [AvroSchemaResolver.java#handleUnionFieldsWithNull](https://github.com/hortonworks/registry/blob/master/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/avro/AvroSchemaResolver.java#L141).

It should be as simple as ensuring that props are carried over from fields during the schema rebuilding in this method. Will update if I have a working solution.

ivangreene:
Resolves #502


ivangreene:
Are any maintainers available to review this? It's quite a simple fix to a problem that makes it unreliable to use props to associate further metadata with a schema

PR #514, opened 2018-11-27
ISSUE-513: Fix storage manager to correctly handle the types in queries

arunmahadevan:
Currently the JDBC storage manager user the value of the query parameter to guess the column type and uses this to construct the prepared statement. This is fragile and breaks when the value appears to be of a different type. For example a varchar column name and query string name="123".

We can leverage the JDBC result set to figure out the sql type and map it to the Schema.Type to fix this.


HeartSaVioR:
+1 on the idea. Could you share how did you test the change? Because there's no test added or being modified. Hard to say it should work well before manually testing against all supported DBMS.

Btw, I'd lean on @satishd and @raju-saravanan reviews and approvals, since it will affect Registry and also might affect SMM as well.

PR #516, opened 2018-12-01
ISSUE-517 Fix concurrent schema version adds.

### LINKED ISSUE #517: Allow multiple concurrent schema version creations by blocking individual requests
When a schema version is to be added we query the database to get the latest schema version's version number, increment the value and it becomes the new schema version's version number. Currently, only one client is allowed to create a version and other concurrent requests are received an error and they should retry again.

raju-saravanan:
When a schema version is to be added we query the database to get the latest schema version's version number, increment the value and it becomes the new schema version's version number. Currently, only one client is allowed to create a version and other concurrent requests are received an error and they should retry again. With this PR, other requests would wait till one of the requests finishes the update. 
A new table "schema_locks" has been exposed as a part of this PR. For every, schema metadata that is added we add a new entry in "schema_locks" table and whenever a schema version is to be added a write lock is obtained on a row corresponding to that schema metadata on "schema_locks" table, then we can exclusively query the database to find the latest schema version. This would resolve the concurrent issue we run into while adding multiple schema versions parallel.

Changes were tested against Mysql, Postgres, Oracle 12c and Oracle 11gr2.

Closes #517

### DIFF
--- a/storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/JdbcStorageManager.java
+++ b/storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/JdbcStorageManager.java
@@ -131,18 +131,22 @@ public class JdbcStorageManager implements TransactionManager, StorageManager {
     }
 
     private boolean getLock(Supplier<Collection<Storable>> supplier, Long time, TimeUnit timeUnit) throws InterruptedException {
-        Long remainingTime = TimeUnit.MILLISECONDS.convert(time, timeUnit);
-        Long startTime = System.currentTimeMillis();
-        while (remainingTime > 0) {
+        long remainingTime = TimeUnit.MILLISECONDS.convert(time, timeUnit);
+
+        if(remainingTime < 0) {
+            throw new IllegalArgumentException("Wait time for obtaining the lock can't be negative");
+        }
+
+        long startTime = System.currentTimeMillis();
+        do {
             Collection<Storable> storables = supplier.get();
             if (storables != null && !storables.isEmpty()) {
                 return true;
             } else {
                 Thread.sleep(500);
-                remainingTime -= (System.currentTimeMillis() - startTime);
-                startTime = System.currentTimeMillis();
             }
-        }
+        } while((System.currentTimeMillis() - startTime) < remainingTime);
+
 
         return false;
     }
--- a/storage/core/src/main/java/com/hortonworks/registries/storage/transaction/TransactionEventListener.java
+++ b/storage/core/src/main/java/com/hortonworks/registries/storage/transaction/TransactionEventListener.java
@@ -107,21 +107,15 @@ public class TransactionEventListener implements ApplicationEventListener {
                 useTransactionForUnitOfWork =
                         unitOfWork.map(UnitOfWork::transactional).orElse(runWithTxnIfNotConfigured);
 
-                TransactionIsolation transactionIsolation = null;
-
-                if (unitOfWork.isPresent()) {
-                    if (unitOfWork.get().transactionIsolation() == TransactionIsolation.APPLICATION_DEFAULT) {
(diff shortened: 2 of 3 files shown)


storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/JdbcStorageManager.java
```diff
@@ -102,6 +104,49 @@ public void update(Storable storable) {
...
+
+        try {
+            return getLock(supplier, time, timeUnit);
+        } catch (InterruptedException e) {
+            throw new StorageException("Failed to obtain a write lock for storable key : " + key);
+        }
+    }
+
+    private boolean getLock(Supplier<Collection<Storable>> supplier, Long time, TimeUnit timeUnit) throws InterruptedException {
+        Long remainingTime = TimeUnit.MILLISECONDS.convert(time, timeUnit);
```
satishd: Check for time value to be always positive and thrown an error if not.

satishd: Declare `remainingTime` and `startTime` as primitives as boxing/unboxing are not required as all usages are primitive.

storage/core/src/main/java/com/hortonworks/registries/storage/impl/jdbc/JdbcStorageManager.java
```diff
@@ -102,6 +104,49 @@ public void update(Storable storable) {
...
+        Long startTime = System.currentTimeMillis();
+        while (remainingTime > 0) {
+            Collection<Storable> storables = supplier.get();
+            if (storables != null && !storables.isEmpty()) {
+                return true;
+            } else {
+                Thread.sleep(500);
+                remainingTime -= (System.currentTimeMillis() - startTime);
+                startTime = System.currentTimeMillis();
```
satishd: This can be simplified by having the below by not setting `startTime` again and not updating `remainingTime`.

```
        long startTime = System.currentTimeMillis();
        do {
            Collection<Storable> storables = supplier.get();
            if (storables != null && !storables.isEmpty()) {
                return true;
            } else {
                Thread.sleep(500);
            }
        } while((System.currentTimeMillis() - startTime) < remainingTime);

```

storage/core/src/main/java/com/hortonworks/registries/storage/transaction/TransactionEventListener.java
```diff
@@ -86,9 +105,24 @@ public void onEvent(RequestEvent event) {
...
                 // get property whether to have unitOfWork with DB default transaction by default
                 useTransactionForUnitOfWork =
-                        unitOfWork.isPresent() ? unitOfWork.get().transactional() : runWithTxnIfNotConfigured;
-                TransactionIsolation transactionIsolation = unitOfWork.map(UnitOfWork::transactionIsolation)
-                                                                      .orElse(TransactionIsolation.DEFAULT);
+                        unitOfWork.map(UnitOfWork::transactional).orElse(runWithTxnIfNotConfigured);
+
+                TransactionIsolation transactionIsolation = null;
+
+                if (unitOfWork.isPresent()) {
```
satishd: This block can be simplified to `Optional.map().orElse()` lambda style. I lean to this style but I am fine with either ways.

```
TransactionIsolation transactionIsolation =
        unitOfWork.map(x -> {
            TransactionIsolation result = x.transactionIsolation();
            if (result == TransactionIsolation.APPLICATION_DEFAULT) {
                result = defaultTransactionIsolation == null ? TransactionIsolation.DATABASE_SENSITIVE : defaultTransactionIsolation;
            }
            return result;
        }
).orElse(TransactionIsolation.DATABASE_SENSITIVE);
```

raju-saravanan:
@satishd: Addressed review comments. Can you have a look?

PR #539, opened 2019-03-04
ISSUE-538 Move transaction package from common to storage-core

### LINKED ISSUE #538: Move transaction package in common to storage-core
`storage-tool` (which depends on `storage-core`) is not only used for `webservice` module in SAM and SR, but also used for independent tool. 

Having `registry-common` as a dependency pulls lots of unnecessary dependencies including `jackson` and `hadoop-client` which are not relevant to storage modules. There's no chance for using transaction package in common without depending on storage-core, so it's safe to move them into `storage-core`.

HeartSaVioR:
This helps us to isolate `storage-core` with `registry-common`, which brings huge number of transitive dependencies especially from `hadoop-client`.

Closes #538

### DIFF
--- a/schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/ConfluentSchemaRegistryCompatibleResource.java
+++ b/schema-registry/rest-service/src/main/java/com/hortonworks/registries/schemaregistry/webservice/ConfluentSchemaRegistryCompatibleResource.java
@@ -20,7 +20,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.hortonworks.registries.common.catalog.CatalogResponse;
 import com.hortonworks.registries.common.ha.LeadershipParticipant;
-import com.hortonworks.registries.common.transaction.UnitOfWork;
+import com.hortonworks.registries.storage.transaction.UnitOfWork;
 import com.hortonworks.registries.common.util.WSUtils;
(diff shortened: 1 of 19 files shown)


HeartSaVioR:
Along with this I would like to discuss regarding maintaining common modules - I'd like to see either of one:

- split the project for maintaining common modules and place only common things (of course, require cross reviews)
- require cross reviews on common modules (at least `common` and `storage-*`) - require each +1 from both side (SAM side, and SR side)

I admit it's an ideal view, but worth to consider. If we are not interested on doing this, please ignore.

HeartSaVioR:
DISCLAIMER: I'm working on shading SR modules from Streamline due to Avro version conflict, and found many modules depend only on `storage-core`. Without this patch we should include `storage-core` to shaded lib and have to deal with transitive dependencies. We will relocate problematic transitive dependencies for sure, but this patch would avoid the huge efforts.

HeartSaVioR:
Now I'm proposing non-shade approach, but I think this issue is still valid.

HeartSaVioR:
@satishd 
Streamline has three usages on `transaction` module. Given that Streamline has to apply the latest Registry version because of #540 and #543, I'd like to address them (include this) altogether.
Please postpone merging this until next release if there's no resource/time for team to reflect the change to SMM.

satishd:
@HeartSaVioR We can merge this PR once @arunmahadevan is also done with his review. We will make respective changes in SMM later once registry is released with these changes.

HeartSaVioR:
This one is approved by @arunmahadevan. #544 is the one @arunmahadevan is not reviewed yet.

Issue #520, opened 2018-12-06
schema-registry-core: produce a separate artifact containing core library only


bitbrain:
# Problem

The project `com.hortonworks.registries:schema-registry-core` is configured to produce a "fat jar" artifact. This means, all required class files are packed into the same .jar file to allow execution of that .jar file.

However, other schema-registry modules depend classes from `schema-registry-core`, even if the intention of these modules is not to directly execute the .jar file. This leads to undesired side effects, since that particular `schema-registry-core` jar file is now on the classpath on any project depending on this schema registry. It contains a lot of additional class files (older Spring class files, Avro class files). This potentially leads to classfile conflicts. Another side effect is that it is not possible to override/remove dependencies coming from the core artifact, since all class files are directly part of the .jar file, instead having them defined as a pom file dependency.

# Solution

Change the Maven shade plugin inside the schema core project to generate multiple artifacts:

- the default artifact should not be executable and should only contain classfiles compiled from the core project. It should produce a lightweight library jar which can be referenced by other projects
- define an additional configuration to generate an executable jar containing all required classfile dependencies to execute it 

*Similar issues:* #223

gcsaba2:
Done

Issue #223, opened 2017-08-08
Schema Client and Schema Serdes un-needed Dependcies


michaelandrepearce:
Currently when bringing into projects Schema Client and the Schema Serdes the transitive dependencies is very large, increasing risk of dependency conflict.

Such dependencies such as org.apache.hadoop:hadoop-client:jar which should simply not be needed by a pure kafka or another other client not on hadoop, should not need, but makes a very large transitive deps.

This seems to be due to a common dependency on registry-common which brings these in, and probably should therefor be split out, or excluded.

This should be reduced to just the core dependencies needed to function, for each component.

michaelandrepearce:
@satishd to do this rather than having to a bunch of dirty exludes, i am thinking of splitting the common module into:

- common-client
- common-server 

where common-server depends on common-client and common-client is as lightweight as possible. wdyt.

satishd:
@michaelandrepearce +1 to make client module having minimal dependencies. Ideally, common-server should not depend upon common-client but both can have dependencies on a single common module. It should also be taken into account that registry-common module is used by http://github.com/hortonworks/streamline project and respective changes to be done there once the changes are done here.

HeartSaVioR:
+1

HdfsFileStorage is clearly server side class so shouldn't be included to `common`, and it's only usage of hadoop-common.

PR #541, opened 2019-03-04
ISSUE-540 Exclude Avro dependency from registry-common module

### LINKED ISSUE #540: Exclude Avro dependency from registry-common module
Having `hadoop-client` as a dependency in `registry-common` brings uncontrolled version of Avro dependencies to too many modules which have registry-common module as transitive dependency. If excluding Avro hurts something, it's a signal to shade&relocate Avro in registry-common to fix it without bothering other modules.

HeartSaVioR:
This would save lots of modules which has `registry-common` as a transitive dependency.

Closes #540

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -73,6 +73,10 @@
                     <groupId>org.eclipse.jetty</groupId>
                     <artifactId>jetty-webapp</artifactId>
                 </exclusion>
+                <exclusion>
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
#223 and #520 look to be food for thought.

HeartSaVioR:
Thanks all for reviewing. I'll push forward to HDF-3.4.0.0 branch as well.

PR #542, opened 2019-03-04
MINOR: Upgrade jackson-databind to v2.9.8 to fix the below CVEs.

kamalcph:
CVE-2018-14719
CVE-2018-14720
CVE-2018-14721
CVE-2018-1000873
CVE-2018-7489
CVE-2018-19362
CVE-2017-15095
CVE-2018-19361
CVE-2017-7525
CVE-2018-19360
CVE-2017-17485
CVE-2018-5968

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -341,7 +341,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <junit.version>4.12</junit.version>
-        <jackson.version>2.9.5</jackson.version>
+        <jackson.version>2.9.8</jackson.version>
         <commons-cli.version>1.3.1</commons-cli.version>
         <commons.version>2.5</commons.version>
         <slf4j.version>1.7.25</slf4j.version>
--- a/schema-registry/serdes/pom.xml
+++ b/schema-registry/serdes/pom.xml
@@ -20,7 +20,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.jaxrs</groupId>
             <artifactId>jackson-jaxrs-json-provider</artifactId>
-            <version>2.7.3</version>
+            <version>${jackson.version}</version>
         </dependency>
         <dependency>
             <groupId>org.glassfish.jersey.media</groupId>


kamalcph:
Tested this branch by running the unit tests and the `KafkaAvroSerDesApp`, `ClickStreamEnrichment` examples.

HeartSaVioR:
Could you add this test into `registry-common` module and see whether it passes the test? The test refers all things in SR registry-common side, so better to have it here. 
If this test fails in SR, it indicates we should fix something in here.

https://github.com/hortonworks/streamline/blob/master/common/src/test/java/com/hortonworks/streamline/common/SchemaTest.java

Below test fails when I upgrade jackson version to 2.9.8 in SAM (I suspect it's regardless of jackson version of SR).

```
@Test
    public void testSerializeNestedSchema() throws Exception {
        ObjectMapper mapper = new ObjectMapper();
        Schema nested = new Schema();
        Schema.Field f1 = new Schema.Field("field1", Type.INTEGER);
        Schema.Field x = new Schema.Field("x", Type.INTEGER);
        Schema.Field y = new Schema.Field("y", Type.INTEGER);
        Schema.Field f2 = Schema.NestedField.of("field2", Arrays.asList(x, y));
        nested.setFields(Arrays.asList(f1, f2));
        String expected = "{\"fields\":[{\"name\":\"field1\",\"type\":\"INTEGER\",\"optional\":false}," +
                "{\"name\":\"field2\",\"type\":\"NESTED\",\"optional\":false,\"fields\":[{\"name\":\"x\",\"type\":\"INTEGER\"," +
                "\"optional\":false},{\"name\":\"y\",\"type\":\"INTEGER\",\"optional\":false}]}]}";
        assertEquals(expected, mapper.writeValueAsString(nested));
        Schema schema2;
        schema2 = mapper.readValue(expected, Schema.class);
        assertEquals(nested, schema2);
    }
```

HeartSaVioR:
Just added into `registry-common` in local dev. and seeing it failed. Would we need to find some necessary changes or there's some known bug?

kamalcph:
@HeartSaVioR 
The test fails in the master branch too.

HeartSaVioR:
@kamalcph @raju-saravanan 

So doesn't it indicate the issue when SR upgrades to jackson-databind 2.9.x? I think the test should be here in SR - see how the test is constructed. It doesn't depend on anything in SAM. It just didn't discovered yet in SR because the test has been placed in wrong place. 

We should do either: find the bug in test and abandon it, or fix it if the test is correct.

cc. @arunmahadevan

arunmahadevan:
Looks like the Schema class in registry would need to be fixed. The result has "type": null in addition to the actual type. I agree this test should be in Registry and we missed to move this one while moving around the classes between registry and SAM.

```
Expected :{"fields":[{"name":"field1","type":"INTEGER","optional":false},{"name":"field2","type":"NESTED","optional":false,"fields":[{"name":"x","type":"INTEGER","optional":false},{"name":"y","type":"INTEGER","optional":false}]}]}
Actual   :{"fields":[{"type":"null","name":"field1","type":"INTEGER","optional":false},{"type":"null","name":"field2","type":"NESTED","optional":false,"fields":[{"type":"null","name":"x","type":"INTEGER","optional":false},{"type":"null","name":"y","type":"INTEGER","optional":false}]}]}
```

HeartSaVioR:
https://github.com/hortonworks/registry/pull/544

PR #547, opened 2019-04-04
ISSUE-548: Make compatible with Avro 1.9.0

### LINKED ISSUE #548: Make compatible with Avro 1.9.0
To provide flexibility to users, we should release a version of Registry which is compatible with both current Avro 1.8.2 and the upcoming version 1.9.0.

The work to get here is mostly in the form of ensuring we are not using methods that have been removed in the new version, and removing usages of shaded or transitive dependencies that were removed.

ivangreene:
Preparing for the upcoming release of Avro 1.9.0,
there are some methods and transitive dependencies
that have been removed.
Resolves #548 

~~This still needs some work regarding default values.
As of 1.8.x, Avro's Schema.Field class exposed
'defaultValue', returning a JsonNode. That method is
package private in 1.9.0-SNAPSHOT, in favor of defaultVal
returning an Object. In this instance, we have no way to
distinguish between 'null' as a default and a field having
no default at all.
So we need to figure out how to manage
these cases. We could try to squeeze in 'hasDefault' before
the new Avro release but we will be binary incompatible at
that point so we would need 2 registry builds for Avro 1.8.x
and 1.9.0.~~

Mostly false alarm on the above comment. I spent a bunch of time
setting up reflective junk to mitigate what I thought was
a problem but it turned out that `defaultVal()` returns
`JsonProperties.NULL_VALUE` when a field has a default
of 'null'. So looks like we are mostly good.
The only failure against Avro 1.9.0-SNAPSHOT came when using the
Confluent schema registry in the schema-registry-webservice
tests.

### DIFF
--- a/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/SchemaBranchKey.java
+++ b/schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/SchemaBranchKey.java
@@ -16,10 +16,10 @@
 
 package com.hortonworks.registries.schemaregistry;
 
-import avro.shaded.com.google.common.base.Preconditions;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 
 import java.io.Serializable;
+import java.util.Objects;
 
 /**
  * This class contains schema branch name and schema metadata name.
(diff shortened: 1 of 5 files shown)


michaelandrepearce:
@ivangreene if a method was previously accessible and now they're removing that method, thats a breaking change in avro, maybe best to discuss in upstream avro project, if they could revert that method or at least keep it but as deperecated, so its not a hard break.

ivangreene:
@michaelandrepearce I will try to see if they will accept a patch to expose this method again. If we can’t find a way to make a registry version compatible with both 1.8 and 1.9 line this will hurt adoption of the new Avro version a lot IMO

ivangreene:
@michaelandrepearce I updated my comment to reflect new findings. So I think we are mostly going to be compatible here aside from the tests using the Confluent registry, that will only need to change when this project wants to start running the CI against 1.9

ivangreene:
@satishd can you have a look at this? Avro 1.9.0 has been released so this will block users from being able to upgrade

ivangreene:
@raju-saravanan I see you just put out 0.8.0 RC1, can you have a look at this PR for inclusion? This will be nice and Avro is planning on releasing 1.9.1 soon so users will want to be able to upgrade

raju-saravanan:
@ivangreene: I am ok with the changes in this PR. I will wait for a day or two for @satishd  and @michaelandrepearce to get their opinion.

PR #579, opened 2019-08-01
Issue 578: Find schema version by fingerprint of schema text.

joaopedroantonio:
Adds a new REST endpoint to find a schema version by the fingerprint of
the schema text.

### DIFF
--- a/schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/ISchemaRegistry.java
+++ b/schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/ISchemaRegistry.java
@@ -83,7 +83,8 @@ public interface ISchemaRegistry extends ISchemaRegistryService {
     SchemaVersionInfo getSchemaVersionInfo(String schemaName, String schemaText, boolean disableCanonicalCheck) throws SchemaNotFoundException, InvalidSchemaException, SchemaBranchNotFoundException;
 
     /**
-     * If there is a version of the schema with the given fingerprint then it returns the respective {@link SchemaVersionInfo}, else it returns null.
+     * If there is a version of the schema with the given fingerprint then it returns the respective {@link SchemaVersionInfo},
+     * else it throws a {@link SchemaNotFoundException}.
      *
      * @param fingerprint MD5 hash of the schemaText
      * @return the {@link SchemaVersionInfo} of the schema version with the given fingerprint.
--- a/schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/ISchemaRegistry.java
+++ b/schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/ISchemaRegistry.java
@@ -83,6 +83,15 @@ public interface ISchemaRegistry extends ISchemaRegistryService {
     SchemaVersionInfo getSchemaVersionInfo(String schemaName, String schemaText, boolean disableCanonicalCheck) throws SchemaNotFoundException, InvalidSchemaException, SchemaBranchNotFoundException;
 
     /**
+     * If there is a version of the schema with the given fingerprint then it returns the respective {@link SchemaVersionInfo}, else it returns null.
+     *
+     * @param fingerprint MD5 hash of the schemaText
+     * @return the {@link SchemaVersionInfo} of the schema version with the given fingerprint.
+     * @throws SchemaNotFoundException when no schema version registered with the given fingerprint.
+     */
+    SchemaVersionInfo findSchemaVersionByFingerprint(String fingerprint) throws SchemaNotFoundException;
+
+    /**
      * @param props properties
      *
      * @return Collects aggregated schema metadata which contains the given properties.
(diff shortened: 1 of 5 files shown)


raju-saravanan:
@joaopedroantonio Thanks for showing interest in the project. May I know the circumstances that you are trying to address with this PR?

joaopedroantonio:
@raju-saravanan Sure, I can give you some context. At DefinedCrowd we're improving our event sourcing infrastructure and as part of that initiative we want to add Schema Registry to store our schema definitions.

For critical services, we want to register new schema versions at deploy time and publish messages without having to contact the Schema Registry in runtime. To accomplish this, instead of using the schema version id in the message, we can use the schema text fingerprint.

This has a couple downsides, the increase in the message size and the (remote) possibility of having a hash collision but it's a trade off we're willing to make for increased resiliency and easier scaling in critical parts of our architecture.

raju-saravanan:
@joaopedroantonio So will you be exposing new ser des protocol in SerDesProtocolHandlerRegistry to use fingerprint to serialize and deserialze message ? We may have a scenario where we have two schema version with same fingerprint esp when two schemas are canonically same but have different default values for certain fields in the schema, in that case with the current implementation - which picks up a first schema from the result set - deserailized message might not be the one a use expects. We can resolve this by sorting the results set and pick up with schema version with latest timestamp. Will you be ok with this ?  @satishd  and @michaelandrepearce can you guys share your opinions on this ?

joaopedroantonio:
@raju-saravanan I wasn't thinking about implementing a new protocol, as we'll be using the Schema Registry in our scope solely to store and manage the schemas. My suggestion is that we keep this PR simply for exposing the REST endpoint and I'll create a new one to implement the new Protocol. What you think?

Regarding the scenario of having two schema version with the same fingerprint, I agree with returning the latest schema version rather than the first or we could even make this endpoint return all of them. I'm find with both options. What you think?

raju-saravanan:
@joaopedroantonio: 
I have answered your question below:
  * Let's come back to implementing a new protocol if somebody else in the community needs it. 
  * For two schemas with the same fingerprint let's return the schema version with the latest timestamp. 

I have couple of more questions, 
  *  Are you planning on exposing this API on schema registry client ?
   * Can you test a write a test case for the API you have exposed ?

joaopedroantonio:
> @joaopedroantonio:
> I have answered your question below:
> 
> * Let's come back to implementing a new protocol if somebody else in the community needs it.
> * For two schemas with the same fingerprint let's return the schema version with the latest timestamp.
> 
> I have couple of more questions,
> 
> * Just to be clear, you will not be using schema registry during serialization and deserialization of messages. The API is only for your reference/debugging purpose to see what is the corresponding schema version for a given fingerprint.
Exactly, we'll handle serialization/deserizalization on our side (at least for now).
> * Are you planning on exposing this API on schema registry client ?
Wasn't planning to but I'll take a look into it and might end up implementing it for the sake of fun. I'll keep it on a separate PR though.
> * Can you test a write a test case for the API you have exposed ?
Yes, I'll do it.

I'll try to update this PR until the end of the week with the two changes we've discussed:
1. Returns the schema version with the latest timestamp.
2. Write a test case for the API exposed.

joaopedroantonio:
@raju-saravanan implemented the changes discussed previously, let me know your thoughts. :)

schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/ISchemaRegistry.java
```diff
@@ -82,6 +82,15 @@ default SchemaVersionInfo getSchemaVersionInfo(String schemaName, String schemaT
      */
     SchemaVersionInfo getSchemaVersionInfo(String schemaName, String schemaText, boolean disableCanonicalCheck) throws SchemaNotFoundException, InvalidSchemaException, SchemaBranchNotFoundException;
 
+    /**
+     * If there is a version of the schema with the given fingerprint then it returns the respective {@link SchemaVersionInfo}, else it returns null.
```
raju-saravanan: Instead of "else it returns null" can we replace it with "else it throws SchemaNotFoundException"?

schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/SchemaVersionLifecycleManager.java
```diff
@@ -616,6 +617,25 @@ public SchemaVersionInfo getSchemaVersionInfo(SchemaVersionKey schemaVersionKey)
...
+        final Collection<SchemaVersionStorable> schemas = storageManager.find(SchemaVersionStorable.NAME_SPACE, queryParams);
+
+        if (schemas.isEmpty()) {
+            throw new SchemaNotFoundException(String.format("No schema found for fingerprint: %s", fingerprint));
+        } else {
+            if (schemas.size() > 1) {
+                LOG.warn(String.format("Multiple schemas found for the same fingerprint: %s", fingerprint));
+            }
+
+            return schemas.stream()
```
raju-saravanan: Instead of sorting the schemas in-memory, can we get sorted schemas from the database and pick the first one? You can have a look at [this](https://github.com/hortonworks/registry/blob/77932b984c32f05cf17957ee0c54b185b302659d/schema-registry/core/src/main/java/com/hortonworks/registries/schemaregistry/DefaultSchemaRegistry.java#L744) code which does this.

raju-saravanan:
@joaopedroantonio : Left few more comments to address.

joaopedroantonio:
Thank you for the comments @raju-saravanan, it think they're addressed. :)

PR #581, opened 2019-08-07
ISSUE 568 : Synchronize Kerberos ticket renewal for non-renewable ticket

### LINKED ISSUE #568: Synchronize Kerberos ticket renewal for non-renewable ticket
Schema registry client uses KerberosLogin class for refreshing Kerberos TGT in a forked thread. For non-renewable ticket we logout() (which destroys the principal along with any private and public creds for that principal) and then login(), if there is a call placed to the schema registry server between this interval schema registry get a 401 Unauthorized Exception. We should block any calls made to the server by the schema registry client between while the TGT is getting refreshed.

raju-saravanan:
Fixes #568

### DIFF
--- a/common-auth/src/main/java/com/hortonworks/registries/auth/KerberosLogin.java
+++ b/common-auth/src/main/java/com/hortonworks/registries/auth/KerberosLogin.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2017 Hortonworks.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
-
- *   http://www.apache.org/licenses/LICENSE-2.0
-
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -30,6 +30,8 @@ import java.util.Date;
 import java.util.Map;
 import java.util.Random;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
 
 /**
  * This class is responsible for logging in to Kerberos and refreshing credentials for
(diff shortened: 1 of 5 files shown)


common-auth/src/main/java/com/hortonworks/registries/auth/KerberosLogin.java
```diff
@@ -246,6 +260,26 @@ private KerberosTicket getTGT() {
         return null;
     }
 
+    private void synchronizeReLogin() throws LoginException {
```
guruchai: I dont see any point having relogin and synchronizeReLogin. Please rename this as relogin and delete the other.

raju-saravanan: The code in try {} block and in else {} block have same lines of code which has been abstracted out as reLogin(). Its better to keep this abstracted out, as synchronizedReLogin() is wrapper over reLogin() which hold the actual logic to reLogin() while synchornizedReLogin() simply provides a synchronization on top of it.

raju-saravanan:
@guruchai: I have exposed NOOPLogin class which will abstract out the login
() / doAction() mechanism in the schema registry.

raju-saravanan:
We can't have reLogin() wrapped in a synchronized block as you have referenced in the previous comment. In the schema registry, we don't have a connection pool, we open up a new connection for every request made, so if we open a new connection between logout() and login() of reLogin() the server will throw 401. Zookeeper and Kafka the connection is reused after the request is over and hence it making a call between logout() and login() works as it still the cookie from the previous call.
     In NOOPLogin where we want to use a plain connection to the server, we don't want a lock for every doAction() call. That is why I have not abstracted out the locking mechanism to its abstract class.

guruchai:
@raju-saravanan 
We can't have reLogin() wrapped in a synchronized block as you have referenced in the previous comment. In the schema registry, we don't have a connection pool, we open up a new connection for every request made, so if we open a new connection between logout() and login() of reLogin() the server will throw 401. 
>> I am asking you to have only the synrchronizedRelogin method instead of both relogin and synrchronizedRelogin methods. You may use the WriteLock. synchronized as you said does not make sense. To me, relogin should be treated like an atomic operation.

Zookeeper and Kafka the connection is reused after the request is over and hence it making a call between logout() and login() works as it still the cookie from the previous call.
>> Connection reuse aside. The point is that we will not be able to make calls between logout and login, for any time, irrespective of presence of cookie or not.

guruchai:
Thanks for the PR!

PR #602, opened 2019-10-07
ISSUE 607 : Schema Registry UI hangs occasionally on first use

### LINKED ISSUE #607: Schema Registry UI hangs occasionally on first use
The schema versions component, uses the "statemachine" API response to show the state of a schema version. This is intermittent issue appearing when referencing the state id before the fetch is complete. The states data will always be available when the schema panel is expanded to view the versions. Hence only for initial loading, the UI state should not break if it takes time to fetch the states information.

joylyn:
Fixes #607

### DIFF
--- a/webservice/src/main/resources/app/scripts/components/SchemaDetail.jsx
+++ b/webservice/src/main/resources/app/scripts/components/SchemaDetail.jsx
@@ -383,7 +383,7 @@ export default class SchemaDetail extends Component{
     const s = schema;
     const {name, evolve} = s.schemaMetadata;
     const currentBranchName = selectedBranch ? selectedBranch.schemaBranch.name : "MASTER";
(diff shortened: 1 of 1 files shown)


guruchai:
@joylyn Can you provide the explanation for this change? I am not able to recreate the issue now inorder to test it.

@kamalcph If it's not too much to ask, Can you apply this patch in your environment and test it? I believe you have a consistent recreate.

joylyn:
The schema versions component, uses the "statemachine" API response to show the state of  a schema version. This is intermitten issue appearing when referencing the state id before the fetch is complete. The states data will always be available when the schema panel is expanded to view the versions. Hence only for initial loading, the UI state should not break if it takes time to fetch the states information.

raju-saravanan:
@joylyn : Going forward lets create Github issues instead of referencing CSP JIRAs directly.

PR #604, opened 2019-10-07
ISSUE-603: Minor: Registry is failing to parse the doAs user

### LINKED ISSUE #603: Minor: Registry is failing to parse the doAs user
There is a premature break in KerberosAuthenticationHandler.java which leads to disregard of doAs user in proxied scenarios.

guruchai:
Fixes #603

### DIFF
--- a/common-auth/src/main/java/com/hortonworks/registries/auth/server/KerberosAuthenticationHandler.java
+++ b/common-auth/src/main/java/com/hortonworks/registries/auth/server/KerberosAuthenticationHandler.java
@@ -450,7 +450,7 @@ public class KerberosAuthenticationHandler implements AuthenticationHandler {
     }
 
     protected static String getDoasUser(HttpServletRequest request) {
-        String doAsUser = null;
+        String doAsUser = "";
         String queryString = request.getQueryString();
         if (queryString != null) {
             String[] pairs = queryString.split(QUERY_STRING_DELIMITER);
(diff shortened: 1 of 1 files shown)


guruchai:
Please merge to CSP-3.0 as well.

common-auth/src/main/java/com/hortonworks/registries/auth/server/KerberosAuthenticationHandler.java
```diff
@@ -461,8 +461,8 @@ protected static String getDoasUser(HttpServletRequest request) {
                         if (doAsUser.isEmpty()) {
                             return null;
                         }
+                        break;
```
raju-saravanan: @guruchai : Will there be multiple "doAs" key in the header ?

raju-saravanan: @guruchai So multiple "doAs" query parameter ?

guruchai: Sorry, with regards to the question, it does not make sense to have multiple doAs query parameters. Since the doAs is typically attached by a proxy server, it makes sense to pick the last one.
So break statement does not make sense. Thanks for pointing!

guruchai:
@raju-saravanan Can you check?

PR #606, opened 2019-10-09
ISSUE 605: Bump up the Avro version to 1.9.1

raju-saravanan:
In Avro 1.8.2 APIs that exposes Jackson classes were deprecated. In the PR https://github.com/hortonworks/registry/pull/547, we moved away from these deprecated APIs to the recommended ones. The PR was then tested against Avro 1.9.0 and but was merged to MASTER with Avro 1.8.2. This PR bumps Avro version from 1.8.2 to 1.9.1 (https://github.com/apache/avro/releases). As part of this update I have commented out confluent tests as confluent serdes still has a dependency on  1.8.2. Will uncomment the tests once confluent updates their dependencies.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -385,7 +385,7 @@
         <kafkaArtifact>kafka_2.11</kafkaArtifact>
         <curator.version>4.0.1</curator.version>
         <curator-test.version>2.12.0</curator-test.version>
-        <avro.version>1.8.2</avro.version>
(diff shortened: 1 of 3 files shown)


guruchai:
Do the confluent tests fail with Avro 1.9.1? If so, that would mean confluent serdes cannot work with SR. If so, i think we should get approval from product on this.

PR #609, opened 2019-10-15
 ISSUE 608 : Support one way SSL in schema registry client.

### LINKED ISSUE #608: Schema Registry Client should support 1 way SSL
Using SchemaRegistryClient, the code that creates the SSL Context requires that a keystore and truststore are both present, even though a keystore should not be required when doing 1-way TLS.

raju-saravanan:
Fixes #608


schema-registry/rest-service/src/test/java/com/hortonworks/registries/schemaregistry/avro/AvroSchemaRegistryClientTest.java
```diff
@@ -96,7 +96,9 @@
 
     @CustomParameterizedRunner.Parameters
     public static Iterable<SchemaRegistryTestProfileType> profiles() {
-        return Arrays.asList(SchemaRegistryTestProfileType.DEFAULT, SchemaRegistryTestProfileType.SSL);
+        return Arrays.asList(SchemaRegistryTestProfileType.DEFAULT,
+                             SchemaRegistryTestProfileType.SSL,
+                             SchemaRegistryTestProfileType.SSL_WITH_SERVER_AUTH);
```
guruchai: nit: SSL_WITH_SERVER_AUTH sounds a bit confusing to me. Do you want to change to it to ONE_WAY_SSL?

raju-saravanan: The profile I introduced requires server authentication but client doesn't and I wanted to add another profile where client requires authentication but server doesn't. If you think its ok to only have the former we can change it to ONE_WAY_SSL, if not I will add SSL_WITH_CLIENT_AUTH profile.

Issue #621, opened 2019-11-13
Fetch invalid schemas 


gmcoringa:
With the new version of schema registry (0.8.1) which updates avro to 1.9.1 its not possible to use schemas that are invalid due to validations included in the new avro version.

One case is when you have an invalid schema and you try to fix it, but the schema-registry try to validate the old version, more information -> https://github.com/hortonworks/registry/issues/617

Another case is using the schema-registry client, for example as the serializer for kafka, because when the client tries to get a schema actually calls addSchemaVersion:

```
java.lang.RuntimeException: An exception was thrown while processing request with message: [Invalid default for field method: null not a "string"]
        at com.hortonworks.registries.schemaregistry.client.SchemaRegistryClient.handleSchemaIdVersionResponse(SchemaRegistryClient.java:591) ~[schema-registry-client-0.5.4.jar:na]
        at com.hortonworks.registries.schemaregistry.client.SchemaRegistryClient.doAddSchemaVersion(SchemaRegistryClient.java:577) ~[schema-registry-client-0.5.4.jar:na]
        at com.hortonworks.registries.schemaregistry.client.SchemaRegistryClient.lambda$addSchemaVersion$1(SchemaRegistryClient.java:515) ~[schema-registry-client-0.5.4.jar:na]
        at com.google.common.cache.LocalCache$LocalManualCache$1.load(LocalCache.java:4739) ~[guava-jdk5-17.0.jar:na]
        at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3524) ~[guava-jdk5-17.0.jar:na]
        at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2317) ~[guava-jdk5-17.0.jar:na]
        at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2280) ~[guava-jdk5-17.0.jar:na]
        at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2195) ~[guava-jdk5-17.0.jar:na]
        ... 176 common frames omitted
```

Note that the stacktrace uses an older version, but the server is updated.

For those who have invalid schemas its not possible to use the new version, unless, all schemas are fixed.

gcsaba2:
Also mentioned in internal jira CDPD-19155

PR #622, opened 2019-11-13
Configuration to disable schema auto creation by the client

gmcoringa:
This pr address https://github.com/hortonworks/registry/issues/621


raju-saravanan:
@gmcoringa Thanks for raising this PR. I have a couple of questions, may I know how this change is going to address your use case? If "schema.auto.create" is disabled then the lastest schema version is being utilized to serialize the data, this would only work if the schemas are backward compatible or fully compatible.

gmcoringa:
@raju-saravanan we have some problems with the latest version when the client tries to get the schema to perform the serialization, because some avro schemas wasn't compatible with the latest avro version (`1.9.1`)

One of them was because the get actually calls `addSchemaVersion`.  This PR tries to address this problem and for us its a good solution because the creation of schemas are done through schema-registry ui with backward compatibility.

Meanwhile we fixed this problem using an older version of schema-registry and fixing all the incompatible avro schemas, but this wasn't enough, because the call to  `addSchemaVersion` perform a validation against all schema versions, due the default validation level `ALL`, which we fixed updating all schemas with the validation level `LATEST`. 

Actually this PR isn't needed anymore, so if you think that do not fit in schema registry design, feel free to close.

PR #629, opened 2019-11-28
ISSUE-631 : Schema Registry numeric properties cannot be set via a property file

heritamas:
During initialization property file is read into a Properties object and
validated against a predefined set of rules. However, it was implicitly
assumed, that property values can only String typed.

This fix adds the necessary conversion beside the already existing validation.

FIxes #629

### DIFF
--- a/schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
+++ b/schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
@@ -1253,6 +1253,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                                       String.class,
                                       "URL of schema registry to which this client connects to. For ex: http://localhost:9090/api/v1",
                                       "http://localhost:9090/api/v1",
+                                      ConfigEntry.StringConverter.get(),
                                       ConfigEntry.NonEmptyStringValidator.get());
 
         /**
@@ -1268,6 +1269,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                                      String.class,
                                      "URL of schema registry to which this client connects to. For ex: http://localhost:9090/api/v1",
                                      DEFAULT_LOCAL_JARS_PATH,
+                                     ConfigEntry.StringConverter.get(),
                                      ConfigEntry.NonEmptyStringValidator.get());
 
         /**
@@ -1289,6 +1291,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                                      Integer.class,
                                      "Maximum size of classloader cache",
                                      DEFAULT_CLASSLOADER_CACHE_SIZE,
+                                     ConfigEntry.IntegerConverter.get(),
                                      ConfigEntry.PositiveNumberValidator.get());
 
         /**
@@ -1300,6 +1303,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                                      Integer.class,
                                      "Expiry interval(in seconds) of an entry in classloader cache",
                                      DEFAULT_CLASSLOADER_CACHE_EXPIRY_INTERVAL_SECS,
+                                     ConfigEntry.IntegerConverter.get(),
                                      ConfigEntry.PositiveNumberValidator.get());
 
         public static final long DEFAULT_SCHEMA_CACHE_SIZE = 1024;
@@ -1313,6 +1317,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                                      Integer.class,
                                      "Maximum size of schema version cache",
                                      DEFAULT_SCHEMA_CACHE_SIZE,
+                                     ConfigEntry.IntegerConverter.get(),
                                      ConfigEntry.PositiveNumberValidator.get());
 
         /**
@@ -1323,6 +1328,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                                      Integer.class,
                                      "Expiry interval(in seconds) of an entry in schema version cache",
                                      DEFAULT_SCHEMA_CACHE_EXPIRY_INTERVAL_SECS,
+                                     ConfigEntry.IntegerConverter.get(),
                                      ConfigEntry.PositiveNumberValidator.get());
 
         /**
@@ -1333,6 +1339,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                                      Integer.class,
                                      "Maximum size of schema metadata cache",
                                      DEFAULT_SCHEMA_CACHE_SIZE,
+                                     ConfigEntry.IntegerConverter.get(),
                                      ConfigEntry.PositiveNumberValidator.get());
 
         /**
@@ -1343,6 +1350,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                                      Integer.class,
                                      "Expiry interval(in seconds) of an entry in schema metadata cache",
                                      DEFAULT_SCHEMA_CACHE_EXPIRY_INTERVAL_SECS,
+                                     ConfigEntry.IntegerConverter.get(),
                                      ConfigEntry.PositiveNumberValidator.get());
 
         /**
@@ -1354,6 +1362,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                                      Integer.class,
                                      "Maximum size of schema text cache",
                                      DEFAULT_SCHEMA_CACHE_SIZE,
+                                     ConfigEntry.IntegerConverter.get(),
                                      ConfigEntry.PositiveNumberValidator.get());
 
         /**
@@ -1364,6 +1373,7 @@ public class SchemaRegistryClient implements ISchemaRegistryClient {
                                      Integer.class,
                                      "Expiry interval(in seconds) of an entry in schema text cache.",
                                      DEFAULT_SCHEMA_CACHE_EXPIRY_INTERVAL_SECS,
+                                     ConfigEntry.IntegerConverter.get(),
                                      ConfigEntry.PositiveNumberValidator.get());
 
         /**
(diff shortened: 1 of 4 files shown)


schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ConfigEntry.java
```diff
@@ -128,6 +137,44 @@ public String toString() {
...
+     *
+     * @param <T> type of the resulting value.
+     */
+    public interface Converter<T> {
+        T convert(String str);
+    }
+
+    public static class IntegerConverter implements Converter<Integer> {
+
+        private static Converter<Integer> instance = new IntegerConverter();
```
raju-saravanan: Can we make this **private static final**  ?

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ConfigEntry.java
```diff
@@ -128,6 +137,44 @@ public String toString() {
...
+        }
+
+        public static Converter<Integer> get() {
+            return instance;
+        }
+    }
+
+    public static class StringConverter implements Converter<String> {
+
+        private static Converter<String> instance = new StringConverter();
```
raju-saravanan: Can we make this **private static final** ?

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ConfigEntry.java
```diff
@@ -128,6 +137,44 @@ public String toString() {
                 '}';
     }
 
+    /**
+     * Parses a String to a typed value
+     *
+     * @param <T> type of the resulting value.
+     */
+    public interface Converter<T> {
+        T convert(String str);
```
raju-saravanan: Maybe we should take Object instead of String here as input for convert(). I believe it makes the usage of convert() cleaner in the SchemaRegistryClient. In schema registry client we won't have to typecast Object into String and then call convert() and we won't have to check if the input is an "instance of" of String and then invoke the convert, all these checks and conversions can be pushed into the convert() function itself.

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ConfigEntry.java
```diff
@@ -128,6 +137,44 @@ public String toString() {
...
+    }
+
+    public static class IntegerConverter implements Converter<Integer> {
+
+        private static Converter<Integer> instance = new IntegerConverter();
+
+        @Override
+        public Integer convert(String str) {
+            return Integer.valueOf(str);
```
raju-saravanan: valueOf can throw NumberFormatException, may be we should wrap this exception with SchemaRegistry custom exception like ConfigTypeConvertionException ?

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ConfigEntry.java
heritamas: Good point, I'm going to change it.

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ConfigEntry.java
heritamas: Good point, I'm going to change it.

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ConfigEntry.java
heritamas: Good point, I'm going to change it.

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ConfigEntry.java
heritamas: I'm going to follow your comment. 
However, I'd like to ask a question. Would it be worth considering typed Property handling, such as Jackson Text dataformats (see: https://github.com/FasterXML/jackson-dataformats-text/tree/master/properties )

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ConfigEntry.java
heritamas: We might do this, but what would be the added value of doing so? We would just exchange one type of exception to an another.

raju-saravanan: The idea is that ConfigTypeConvertionException - or something with a better name - would act as a root exception class for Converters, meaning that it would not only wrap NumberFormatException mentioned here but it could be used for other exceptions like in the IntegerConverter I have commented on before. By having a single root exception class it would be easier to debug where the issue is coming from.

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
```diff
@@ -1253,6 +1253,7 @@ public String run() {
                                       String.class,
                                       "URL of schema registry to which this client connects to. For ex: http://localhost:9090/api/v1",
                                       "http://localhost:9090/api/v1",
+                                      ConfigEntry.StringConverter.get(),
```
kamalcph: The type of the ConfigEntry is already defined. Can we implicitly choose the converter instead of passing it in constructor?

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/ConfigEntry.java
```diff
@@ -128,6 +137,48 @@ public String toString() {
...
+    public static class IntegerConverter implements Converter<Integer> {
+
+        private static final Converter<Integer> instance = new IntegerConverter();
+
+        @Override
+        public Integer convert(Object obj) {
+            if (obj instanceof String) {
+                return Integer.valueOf((String)obj);
+            } else {
+                return ((Integer) obj);
```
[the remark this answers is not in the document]
kamalcph: Can we also log the type of the object?
`Found + obj.getClass() + type instead of integer with value : " +obj`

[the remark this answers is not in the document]
heritamas: I've added this specific Exception, also taking Kamal's comment into account (give info about the type as well)

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
heritamas: Is it possible? I mean, in an elegant way.
Once a `ConfigEntry<T>` is created, we know its type parameter `T`, so in case, for instance,  of a `String` a value of type `Converter<String>` can be assigned to a variable typed `Converter<? extends String>`.

But what about the general case? We should construct an expression whose result can be either `Converter<Integer>` or `Converter<String>` and assign to (or return as) something typed `Converter<? extends something>`. The common ancestor of `String` and `Integer` is only `Object`. In this case we would have a `Converter<Object>`, which a solution to this problem - but not a useful one. I think, it is cleaner to stick to the way as, for example, a `Validator` is chosen. All the information we need is present at compile time.

kamalcph: You can refer this method (https://github.com/apache/kafka/blob/eaccb9292984455a2978a635a224fb51f3b051b9/clients/src/main/java/org/apache/kafka/common/config/ConfigDef.java#L633) which parses the value based on it's type.

schema-registry/common/src/main/java/com/hortonworks/registries/schemaregistry/errors/ConfigTypeConversionException.java
```diff
@@ -0,0 +1,8 @@
+package com.hortonworks.registries.schemaregistry.errors;
```
raju-saravanan: Can you add the Cloudera license header here?

heritamas: Yes, of course - I guess, it is the same as for other files.

raju-saravanan: Yup, same as the other files.

kamalcph:
@heritamas @raju-saravanan 

I usually avoid taking extra fields if it can be inferred from other fields. Can we avoid using the `Converter` and use the `Type` field instead? I was thinking something like below in ConfigEntry class:

```java
    @SuppressWarnings("unchecked")
    public static <T> T parseValue(Class<T> clazz, Object value) {
        if (clazz.equals(String.class)) {
            return (T) String.valueOf(value);
        } else if (clazz.equals(Integer.class)) {
            try {
                return (T) Integer.valueOf(value.toString());
            } catch (NumberFormatException ex) {
                throw new ConfigTypeConversionException(ex.getMessage(), ex);
            }
        } else if (clazz.equals(Long.class)) {
            return (T) Long.valueOf(value.toString());
        } else {
            throw new ConfigTypeConversionException("Found " + value.getClass() + " instead of " + clazz + " for value: " + value);
        }
    }

finalValue = ConfigEntry.parseValue(configEntry.type(), value);
```

heritamas:
@kamalcph , regarding your comment:
I feel, whatever approach we choose, it would not be _elegant_, or clean - class casting would remain somewhere in the code.

Would it worth to refactor Property handling to a typed solution, like [this](https://github.com/FasterXML/jackson-dataformats-text/tree/master/properties )?

kamalcph:
Adding a new dependency may not worth it. Let's go with your approach.

raju-saravanan:
@heritamas and @kamalcph  I gonna go ahead and merge it PR.

PR #638, opened 2019-12-06
SR kerberos principle support is added

vomoshkovskyi:


webservice/src/main/java/com/hortonworks/registries/webservice/RegistryConfiguration.java
```diff
@@ -53,6 +50,9 @@
     private String httpProxyUsername;
     private String httpProxyPassword;
 
+    @JsonProperty
+    private AuthMethodConfiguration authenticationMethod;
```
vomoshkovskyi: @raju-saravanan @guruchai We need to decide the correct name for this configuration.

bin/registry-server-start.sh
```diff
@@ -47,6 +47,8 @@ if [ ! "x$EXT_CLASSPATH" = "x" ]; then
  CLASSPATH=$CLASSPATH:$EXT_CLASSPATH;
 fi
 
+CLASSPATH=$CLASSPATH:$base_dir/hadoop-conf/
```
vomoshkovskyi: core-site.xml should be copied to this folder during SR installation. _core-site.xml_ should be in the classpath for correct UGI initialization.

guruchai:
@vomoshkovskyi I thought on this further. Here us my opinion:
Authentication
1. The authorization does not work if core-site.xml is not available. core-site.xml is mandatory for hadoop based services. Schema Registry is not one of them.

Location of Authentication
1. What happens if we have different form of authentication for Ranger in the future. Let's say BASIC Authentication based on LDAP credentials. We should not touch the core classes like RegistryApplication.java everytime there's a new authentication support for Ranger or any other authorizer.

The idea of identity comes into the picture when an entity has to prove itself to a different service. This is an attribute of a client. And so i think it should be ideally on the client side of the code. In this case, it should be on the plugin side. That way, you could extend many authentication mechanisms. And all the plugin configuration can be done in a separate config file without having to touch core service. In my opinion, KafkaRangerPlugin is good approach.

vomoshkovskyi:
@guruchai 
**Authentication**
Authentication will work but it will be simple, not Kerberos.

1. Location of Authentication
The name "AuthenticacationMethod" maybe not the best. We just need a configuration section to specify the Kerberos principal for SR. 

I propose to set a meeting and discuss this. In my opinion, SR, like all the other services should have its own principle. It can be used to communicate with the database server.

common/src/main/java/com/hortonworks/registries/common/AuthMethodConfiguration.java
```diff
@@ -0,0 +1,40 @@
+package com.hortonworks.registries.common;
```
guruchai: Cloudera licence is missing/.

vomoshkovskyi:
The feature was merged as a part of another PR.

PR #647, opened 2020-01-09
ISSUE-646 basic auth in client

bpgergo:


schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
```diff
@@ -1397,6 +1404,26 @@ public String run() {
          */
         public static final int DEFAULT_READ_TIMEOUT = 30 * 1000;
 
+        /**
+         * Username for basic authentication.
+         */
+        public static final ConfigEntry<String> BASIC_AUTH_USERNAME =
+                ConfigEntry.optional("schema.registry.client.basic.auth.username",
```
guruchai: minor: I believe clients need not know that basic auth lies underneath. So "schema.registry.auth.username" should be better.

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
```diff
@@ -1397,6 +1404,26 @@ public String run() {
...
+                        String.class,
+                        "Username for basic authentication",
+                        null,
+                        ConfigEntry.NonEmptyStringValidator.get());
+
+        /**
+         * Password for basic authentication.
+         */
+        public static final ConfigEntry<String> BASIC_AUTH_PASSWORD =
```
guruchai: minor: same as above.

bpgergo: fixed

schema-registry/client/src/main/java/com/hortonworks/registries/schemaregistry/client/SchemaRegistryClient.java
bpgergo: fixed

bpgergo:
@raju-saravanan 
re: Do you see any value in exposing other modes of authentication apart from BASIC here ?

BASIC mode was needed for accessing SR through a Knox proxy, I just wanted to add a minimalistic patch to solve this issue. 

I am not sure about any specific use case when other auth modes would be needed, of course there is possibility that end users would find other modes valuable as well.

raju-saravanan:
@bpgergo : Thanks for the clarification

PR #655, opened 2020-01-22
CSP-831: Delete schema metadata when invalid schema is added

joylyn:


raju-saravanan:
@joylyn: Please create a Github issue for this and reference that in the PR.

PR #663, opened 2020-02-06
ISSUE 640: Fix URL failover mechanism in Schema Registry Client

### LINKED ISSUE #640: SchemaRegistryClient doesn't take advantage of the UrlSelector for failover scenarios
The SchemaRegistryClient class constructs a UrlSelector but doesn't catch connection exceptions to call urlWithError and select and new URL in the case where a call to a URL fails.

raju-saravanan:
Fixes #640


raju-saravanan:
FYI, I am planning to add a few more test cases. Will update the PR once its done

PR #668, opened 2020-02-14
MINOR: possibility to include mysql-connector jar in final tar.gz

akatona84:
This functionality is tied to `with-mysql-jar` profile

### DIFF
--- a/registry-dist/pom.xml
+++ b/registry-dist/pom.xml
@@ -80,23 +80,6 @@
                     <version>5.1.40</version>
                 </dependency>
             </dependencies>
-            <build>
-                <plugins>
-                    <plugin>
-                        <artifactId>maven-dependency-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


raju-saravanan:
@akatona84: May I know the motivation behind this ?

akatona84:
> @akatona84: May I know the motivation behind this ?

Yes, it is required for cdp runtime. Libs dir won't be writeable by schemaregistry (otherwise bootstrap would download this jar)

raju-saravanan:
@akatona84: Alright. How will the oracle ojdbc dependencies be dealt with in CDP runtime?

akatona84:
> @akatona84: Alright. How will the oracle ojdbc dependencies be dealt with in CDP runtime?

I think it wasn't provided in csp either. User will copy them in place.

PR #672, opened 2020-02-20
ISSUE 665 : Ensure SR cache is invalidated properly in secure mode when running in HA

### LINKED ISSUE #665: Ensure SR cache is invalidated properly in secure mode when running in HA
Schema registry server caches entities like metadata, schema version, etc. In HA mode - where any of the instances can serve the request - the server that served the request would leave the rest of the instance's internal cache out of sync. Schema registry has an internal mechanism to invalidate the caches of the remaining instances, but it is not security-aware.

raju-saravanan:
Fixes #665 

The PR is yet to be tested in Kerberos mode.


raju-saravanan:
Drop this approach

PR #673, opened 2020-02-24
CSP-823 : Disable caching on server side

raju-saravanan:
In order to support Schema Registry HA, disabled caching on the server side.


raju-saravanan:
@guruchai, please have a look at this when you get time.

PR #679, opened 2020-03-05
ISSUE-678: UI does not show parse error for non-breaking space when adding a schema

joylyn:


raju-saravanan:
Instead of showing "schema is invalid". Can you also show what is the issues seen with the schema and on what lines do you have the issues in?

image

joylyn:
parse_error
I have updated the code to show the parse error returned.
@raju-saravanan

PR #700, opened 2020-04-01
Introduce HTTP header authentication

### LINKED ISSUE #703: Restrict access from K8s ingresses using HTTP header authentication
Background:
When using Registry in a K8s environment, it is possible to expose the service through different ingresses. However, not all ingresses should be allowed to change and/or use some of the API.

Acceptance criteria:
GIVEN a registry deployed in a protected environment and exposed through external reverse proxies
WHEN an external user is reaching the service through an external name
THEN the registry should allow only certain HTTP operations and not others
AND the registry administrator should be able to configure what is and isn't allowed

lucamilanesio:
ISSUE-703: Introduce HTTP header authentication
    
Allow to select read-only and read/write users based on an
arbitrary HTTP header.
    
This enable the simplest integration with K8s ingress restrictions
where traffic coming from the outside of the cluster should be allowed
to change any entry in the registry.
    
Closes #703

### DIFF
--- a/common-auth/src/main/java/com/hortonworks/registries/auth/server/HttpHeaderAuthenticationHandler.java
+++ b/common-auth/src/main/java/com/hortonworks/registries/auth/server/HttpHeaderAuthenticationHandler.java
@@ -1,0 +1,259 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License. See accompanying LICENSE file.
+ */
+package com.hortonworks.registries.auth.server;
+
+import com.hortonworks.registries.auth.client.AuthenticationException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.*;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * The <code>HttpHeaderAuthenticationHandler</code> provides a http authentication mechanism that restricts
+ * the read and write operations to input HTTP header/values received.
+ * <p>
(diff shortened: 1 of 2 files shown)


gcsaba2:
Hi Luca, I've looked at your PR and it looks good. My question is, does it make sense to allow _any_ header to be defined?

The way I see it, with this we could set "readonly.header.name" to the value of "Color" and then we'd require the server to send such a header with some predefined values, like "Color: blue". Is there a use case for this?

What if we restrict the headers to a set of predefined values, like Host, User, User-Agent, Cookie? Do you think it makes sense?

lucamilanesio:
Hi @gcsaba2 thanks for looking at my PR and providing your feedback on it.
See my comments below.

Kind Regards.
Luca.


> Hi Luca, I've looked at your PR and it looks good. My question is, does it make sense to allow _any_ header to be defined?

There is no advantage in reducing the possible headers and we would lose flexibility in terms of use-cases.

The typical use-case (my one) is to restrict the access for different K8s ingresses, and thus using the Host header. However, I do know that sometimes some SSO systems have some custom X-Something header for authenticated calls that could be used with this authentication mechanism.

> The way I see it, with this we could set "readonly.header.name" to the value of "Color" and then we'd require the server to send such a header with some predefined values, like "Color: blue". Is there a use case for this?

Color, possibly not :-) but `X-SSO-Role` yes (I made `X-SSO-Role` up for this example, but yes, there could be some other use-case). 

> What if we restrict the headers to a set of predefined values, like Host, User, User-Agent, Cookie? Do you think it makes sense?

Do you see any advantage in doing that?

gcsaba2:
You're right about the X-headers. I can't think of any advantage of restricting the list of headers, other than the old rule in security that it's better to restrict as much as possible :)

I'll try to think of ways how to break this code and will get back to you next week. If no one has any objection I think we can merge this. Thank you for your contribution!

One scenario that popped in my mind. Java regex is always unreliable because you don't know if it's greedy or not. Let's say you want to restrict POST method to just the readwrite users and everything else to readonly.

You would have something like this:

readonly.resources: "(GET|PUT|POST|DELETE) .*"
readwrite.resources: "(POST) .*"

Because there is no ^ at the beginning of the regex, we don't necessarily match from the start. So as a hacker with only "readonly" permissions, I could send this request:

POST /dropAllTables?GET%20/someSafePage

The string after the ? would be matched and the user would be permitted to enter the website. The HTTP server would interpret the request correctly as a POST request and would delete all tables as requested.

Small java code to demonstrate this:

```
String requestHdr = "POST /dropAllTables?GET%20/someSafePage";
Pattern regex = Pattern.compile("(GET|PUT|POST|DELETE) .*");
System.out.println("Is this req allowed? " + regex.matcher(requestHdr).matches());  // true
```

Let me know if you see this scenario as a possibility. I know we're talking about kubernetes which is usually inside an intranet, but it's better to be safe than sorry.

Issue #705, opened 2020-04-06
Revert the Jetty version override


viktorsomogyi:
It was discovered that overriding the Jetty version will result in a bug when SR is used in a secure environment:
```
java.lang.IllegalStateException: KeyStores with multiple certificates are not supported on the base class org.eclipse.jetty.util.ssl.SslContextFactory. (Use org.eclipse.jetty.util.ssl.SslContextFactory$Server or org.eclipse.jetty.util.ssl.SslContextFactory$Client instead)
        at org.eclipse.jetty.util.ssl.SslContextFactory.newSniX509ExtendedKeyManager(SslContextFactory.java:1275)
        at org.eclipse.jetty.util.ssl.SslContextFactory.getKeyManagers(SslContextFactory.java:1256)
        at org.eclipse.jetty.util.ssl.SslContextFactory.load(SslContextFactory.java:374)
        at org.eclipse.jetty.util.ssl.SslContextFactory.doStart(SslContextFactory.java:245)
        at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
        at org.eclipse.jetty.util.component.ContainerLifeCycle.start(ContainerLifeCycle.java:169)
        at org.eclipse.jetty.server.Server.start(Server.java:407)
        at org.eclipse.jetty.util.component.ContainerLifeCycle.doStart(ContainerLifeCycle.java:117)
        at org.eclipse.jetty.server.handler.AbstractHandler.doStart(AbstractHandler.java:97)
        at org.eclipse.jetty.server.Server.doStart(Server.java:371)
        at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
        at io.dropwizard.cli.ServerCommand.run(ServerCommand.java:53)
        at io.dropwizard.cli.EnvironmentCommand.run(EnvironmentCommand.java:44)
        at io.dropwizard.cli.ConfiguredCommand.run(ConfiguredCommand.java:87)
        at io.dropwizard.cli.Cli.run(Cli.java:78)
        at io.dropwizard.Application.run(Application.java:93)
        at com.hortonworks.smm.kafka.webservice.KafkaAdminApplication.main(KafkaAdminApplication.java:103)
```
The real solution here will be to upgrade the Dropwizard version.

PR #706, opened 2020-04-06
ISSUE-705: Revert the Jetty version overrides

viktorsomogyi:
Addresses issue #705


akatona84:
It should have an issue, to explain why to revert, about wildcard certs and the exception.

akatona84:
Okok, there's the issue attached, ny mistake

PR #728, opened 2020-05-25
[CDPD-12578] merge to public repo 0.9

gcsaba2:
Upgrading the public repo with the latest changes, after that we'll cut the 0.9 branch


akatona84:
I think we can skip synching stuff from ci directory.

PR #766, opened 2022-10-13
ISSUE-765: Standardize copyright/license headers

acsaki:


acsaki:
Closed since headers don't need to be updated, it turns out.

PR #777, opened 2022-11-07
ISSUE-776 SchemaVersionProtocolHandlerTest is flaky

acsaki:


dorwi (changes requested):
We should trow an exception even when the input protocol version is not a byte. So have a test with an input ((int)260) -> we expect an AvroException, but it wont be thrown since internally this will be converted to ((byte) 4)

Let's fix the byteValue function in AbstractAvroSnapshotSerializer

